PR #1, opened 2021-08-20
Candidate

biwiki:
@aregtech still missing some commits, but please could you help to check this?


aregtech (approved):
Approve with note:
some files are old or deprecated. They will be removed after merge.

aregtech:
@biwiki , some files where either old or deprecated. I'll remove them after merge. Please check the development file. In particular, the generate_files.bat / generate_files.sh files are renamed to generate.bat / generate.sh

PR #41, opened 2022-02-05
Integrate CMake file

Ali-Nasrolahi:
# Brief description on changes

Root cmake created with initializing aliases.

configuration files placed in conf/cmake. some of makefile features which represented in conf/make are not compatible with cmake sysntax, therefore need a little bit polishing. (for instance configuring the platform in user.cmake )

Absolutely open to any changes or resolving any mistake :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 3.18.0)
+
+# Areg root directory
+set(CMAKE_ROOT_FILE "${CMAKE_SOURCE_DIR}")
+set(AREG_SDK_ROOT "${CMAKE_ROOT_FILE}")
+set(CMAKE_CONFIG_DIR "${AREG_SDK_ROOT}/conf/cmake")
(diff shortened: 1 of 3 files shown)


aregtech (approved):

aregtech:
Thank you, @Ali-Nasrolahi !
It's a good start. Would be nice if you remove multiple new-lines at the end or middle of files. Comments / descriptions are always welcomed.

Ali-Nasrolahi:
Definitely. will be fixed on next PR.

biwiki:
Sorry for jumping in the middle!
If this new build system can work with earlier versions of cmake than 3.18.0 then it would be great that we can relax the rule on the minimum cmake version.

Ali-Nasrolahi:
I don't assume any specific feature which needs exactly 3.18 version, though this is just for the sake of compatibility with my own local system.

biwiki:
Thanks! So I am supposed that it is okay to update it to 3.16 as it is the one I have here:
https://github.com/aregtech/areg-sdk/pull/42

PR #42, opened 2022-02-06
cmake minimum version

biwiki:
Relaxing the rule on minimum cmake version


aregtech:
@biwiki, the 3.16 should be fine. Any switch to a newer version should have good reason.
Thank you for your update! I'll review and merge.

aregtech (approved):

PR #43, opened 2022-02-07
NESocket - Use std::string instead of String

biwiki:
=== NOT READY TO MERGE ===
Converting String to std::string in the areg base can be done incrementally like this.
Still not compilable, but it is quite promising that we can move to std::string step by step.

### DIFF
--- a/framework/areg/base/private/SocketClient.cpp
+++ b/framework/areg/base/private/SocketClient.cpp
@@ -37,10 +37,10 @@ bool SocketClient::createSocket( void )
 
     if ( mAddress.isValid() )
     {
-    	SOCKETHANDLE hSocket = NESocket::clientSocketConnect(static_cast<const char *>(mAddress.getHostAddress()), mAddress.getHostPort());
+        SOCKETHANDLE hSocket = NESocket::clientSocketConnect(mAddress.getHostAddress().c_str(), mAddress.getHostPort());
         if ( hSocket != NESocket::InvalidSocketHandle )
         {
-        	mSocket = std::make_shared<SOCKETHANDLE>(hSocket);
+            mSocket = std::make_shared<SOCKETHANDLE>(hSocket);
         }
     }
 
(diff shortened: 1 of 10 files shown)


framework/areg/base/private/SocketClient.cpp
```diff
@@ -37,10 +37,10 @@ bool SocketClient::createSocket( void )
 
     if ( mAddress.isValid() )
     {
-    	SOCKETHANDLE hSocket = NESocket::clientSocketConnect(static_cast<const char *>(mAddress.getHostAddress()), mAddress.getHostPort());
+        SOCKETHANDLE hSocket = NESocket::clientSocketConnect(mAddress.getHostAddress().c_str(), mAddress.getHostPort());
```
aregtech: This is OK. But when we complete converting to `std::string`, we'll need to convert `const char *` in function parameters into `std::string_view` or `const std::string &`. 
At the moment it is fine and I leave like it is.

aregtech:
@biwiki, thank you for changes.
The MS Build fails. The line 227 in NESocket.cpp file --> please change `.isEmpty()` ==> `.empty()`.
It is MS VS specific code and do not see when compile with GCC.

aregtech (changes requested):
Please check MS Build.
File NESocket.cpp, line 227 is MS VS specific code --> `isEmpty()` method does not exist.

biwiki:
@aregtech Yes, this PR breaks the build, thus it is not ready to merge.
In fact, this is a tiny demonstration for the approach that we can use to convert String to std::string.

framework/areg/base/private/SocketClient.cpp
biwiki: ACKed

biwiki:
@aregtech please take another look at this PR, thanks!

aregtech (approved):
Thank you!

PR #44, opened 2022-02-13
Thread/ThreadAddress - Use std::string instead of String

biwiki:
First commit for Thread/ThreadAddress in areg-base for converting from String to std::string

### DIFF
--- a/framework/areg/base/ThreadAddress.hpp
+++ b/framework/areg/base/ThreadAddress.hpp
@@ -261,6 +261,7 @@ inline std::string ThreadAddress::convToString(void) const
 inline const IEInStream & operator >> (const IEInStream & stream, ThreadAddress & input)
 {
     //FIXME
+    // return ( stream >> input.mThreadName );
     String name;
     stream >> name;
     input.mThreadName = std::string(name);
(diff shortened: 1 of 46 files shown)


aregtech:
@biwiki, MSBuild fails. Can you please see the logs?

biwiki:
> @biwiki, MSBuild fails. Can you please see the logs?

yep, I will check the errors

framework/areg/base/ThreadAddress.hpp
```diff
@@ -260,10 +260,16 @@ inline String ThreadAddress::convToString(void) const
 //////////////////////////////////////////////////////////////////////////
 inline const IEInStream & operator >> (const IEInStream & stream, ThreadAddress & input)
 {
-    return ( stream >> input.mThreadName );
+    //FIXME
```
biwiki, line 263: This still needs some improvements, I think

aregtech: Hi @biwiki 
Can you please instead of `//FIXME` write
```cpp
//FIXME
// return ( stream >> input.mThreadName );
```
That I know what exactly it was and what exactly should be fixed.

biwiki: It is fixed now, thanks @aregtech

aregtech (approved):
Thank you, @biwiki

PR #46, opened 2022-02-20
failed to initialize with substr

biwiki:
Couple of bug fixes

### DIFF
--- a/framework/areg/base/private/Process.cpp
+++ b/framework/areg/base/private/Process.cpp
@@ -16,6 +16,7 @@
 #include "areg/base/Process.hpp"
 
 #include "areg/base/File.hpp"
+#include <filesystem>
 
 //////////////////////////////////////////////////////////////////////////
 // Process class implementation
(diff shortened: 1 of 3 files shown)


framework/areg/base/private/Process.cpp
```diff
@@ -44,14 +44,17 @@ void Process::_initPaths( const char * fullPath )
     if ( pos != std::string::npos )
     {
         mProcessPath   = mProcessFullPath.substr(0, pos);
-        mProcessName   = mProcessFullPath.substr(pos + 1);
+        // looks weird, but substr does not work without <str>
+        std::string str = mProcessFullPath.substr(pos + 1);
```
aregtech: @biwiki, if we change the type of  `mProcessName` and other members from `String` to `std::string`, there will be no need for intermediate variable, which has impact on speed and memory (defragmentation). Can you please change the type of object members?

framework/areg/base/private/Process.cpp
```diff
@@ -38,20 +39,11 @@ Process::Process( void )
...
-    if ( pos != std::string::npos )
-    {
-        mAppName    = mProcessName.substr(0, pos);
-        mProcessExt = mProcessName.substr(pos + 1);
-    }
+    mProcessFullPath = std::filesystem::absolute(fullProcessPath).c_str();
+    mProcessPath = std::filesystem::absolute(fullProcessPath).parent_path().c_str();
+    mProcessName = std::filesystem::absolute(fullProcessPath).filename().c_str();
+    mAppName     = std::filesystem::absolute(fullProcessPath).stem().c_str();
+    mProcessExt  = std::filesystem::absolute(fullProcessPath).extension().c_str();
```
aregtech: I think we can optimize here, because `mProcessFullPath = std::filesystem::absolute(fullProcessPath)` and there is no need to re-calculate the full path.
@biwiki, can you please fix this?

aregtech (changes requested):
@biwiki , please have a look at comment.

aregtech:
The MS Build failed.

framework/areg/base/private/Process.cpp
biwiki: Yep, I still have something to do with this method, build is failing for Windows and I will fix it

biwiki: Now I think this class should hold an instance of std::filesystem::path for the full process path. Other information can be extracted at runtime via C++17 filesystem. Different string members in the current class declaration is kind of duplicated info.

aregtech:
I'll make some changes later. One test build fails because of MFC dependencies.

PR #49, opened 2022-03-05
Convert String to std::string - cont

biwiki:


biwiki:
I am thinking to disable the compilation of examples/17_winchat on this branch to speed up the development a bit, at the end we will revisit this example program.

aregtech:
> 

We can for a while disable 17_winchat.

framework/areg/base/private/Version.cpp
```diff
@@ -87,15 +89,24 @@ Version & Version::convFromString( const char * version )
...
-    mPatch  = (mMajor != 0) && (mMinor != 0) ? patch.convToUInt32() : 0;
+    auto pos2 = temp.find_last_of(NECommon::OBJECT_SEPARATOR);
+    if (pos2 != std::string::npos && pos2 != pos1)
+    {
+        const std::string minor = temp.substr(pos1 + 1, pos2 - pos1 - 1);
+        mMinor = std::stoi(minor);
+
+        const std::string patch = temp.substr(pos2+1);
+        mPatch  = (mMajor != 0) && (mMinor != 0) ? std::stoi(patch) : 0;
+    }
```
aregtech, line 103: @biwiki
please double check the possible errors like:
"1", "1.", "1.9", "1.9.", ".9", ".9.", "1.2.3.4.5", etc. How the version is converted here?

framework/areg/base/private/Version.cpp
```diff
@@ -87,15 +89,24 @@ Version & Version::convFromString( const char * version )
     mMinor  = 0;
```
aregtech, line 90: The method ` Version & Version::convFromString( const char * version )` can be converted ` Version & Version::convFromString( const std::string & version )` 
What do think?

aregtech (changes requested):
@biwiki , can you please check the error cases? As well please see my comment regarding interface changes. I think, everywhere where we have `const char *` as a parameter and if in the body of function it is converted to `std::string` we can change the interface to pass parameter as `const std::string &`.

framework/areg/base/private/Version.cpp
biwiki: I thought about using std::regex here, and now I really think that I should go for it

framework/areg/base/private/Version.cpp
biwiki: The plan is to go through entire library to review the usage of const char* later on after all usages of String are converted to std::string

framework/areg/base/private/Version.cpp
aregtech: @biwiki, not sure about regex. Let's keep as simple as possible and use as less resources as possible. All what we do we'll need to adapt to RTOS. Even for C++17 not sure, because some RTOS support only C++11.

aregtech (approved):
Thank you!

PR #52, opened 2022-04-02
20220205 issue18 cmake

Ali-Nasrolahi:
Summery of changes:

- Comments added to `/conf/cmake/` configuration files.
- Some improvements in and replacement in cmake configs.
- New file - `pre-project.cmake` - added for some settings which must be added **before** `project()`.

In this phase, I must say some problems might be encountered during cross-compiling for specific scenarios, though I'm trying to solve them.

Open to any modification or suggestion :)


framework/CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ include("${AREG_BASE}/areg/CMakeLists.txt")
 include("${AREG_BASE}/mcrouter/CMakeLists.txt")
 
 
-# Adding  source files
+# Adding  source files to a single executable.
 add_executable(${PROJECT_NAME} ${areg_SRC} ${mcrouter_SRC})
```
biwiki: Later on, we may need to change the executable name to something like "mcrouter"${AREG_BIN_EXT}
And this target needs to be broken into 2 steps as well:
 - ${areg_SRC} should be built into a library first, this can be a static or a shared lib based on the user input
 - mcrouter then links to the library

Ali-Nasrolahi: Thanks @biwiki.
 **Noted**.
I need to tweak *cmake/common.cmake* and *cmake/user.cmake* a little bit more, then it'll be ready to create multiple targets to manage its behavior as you described. (In the next PR this issue will be addressed).

aregtech: Thank you @Ali-Nasrolahi for changes and @biwiki for the review

Good point, @biwiki. In general

1. At the moment the sources support only Windows and Linux. It is planned to support other OS, including RTOS. The target should be specified in a user specific CMake file located in `conf/cmake` folder.
2. The areg library should be compiled as shared or static, which is compiled first. The option is set in user config. 
3. We should consider the case that Windows and Linux have different extensions. At the moment the extensions are EXE or OUT. But different projects may have different policies. For example, some may prefer do no have extension at all and we should give this possibility.

There is a one more thing to keep in mind, it is related with the SIML service interface description files (see in `examples`). The idea is that these files are added in the project (in CMake). The system should automatically call code generator, generate source files and create a static library, which will be added in the project. Right now, either in `make`, `Eclipse` project or `MS Visual Studio` project files these are done manually. But we'll need to change. The details of these I'll explain later.

aregtech (changes requested):
@Ali-Nasrolahi, can you please have a look at the comments?

Ali-Nasrolahi:
@aregtech and @biwiki please  review new modification.
Is this what you had in mind?

biwiki:
> @aregtech and @biwiki please review new modification. Is this what you had in mind?

Nearly perfect :-)
I tried to compile mcrouter with shared lib option, and the shared lib is outputed to:
areg-sdk/product/build/OFF/UNIX-x86_64-Release/lib/libareg.so
Please check if that is the expected path.

framework/CMakeLists.txt
```diff
@@ -2,9 +2,18 @@ include("${AREG_BASE}/areg/CMakeLists.txt")
...
+else()
+    add_library(areg SHARED ${areg_SRC} ${mcrouter_SRC})
+endif()
+# Library configs
+set_property(TARGET areg PROPERTY CXX_STANDARD 17)
+set_property(TARGET areg PROPERTY LIBRARY_OUTPUT_DIRECTORY ${ProLibDir})
+
 # Adding  source files to a single executable.
-add_executable(${PROJECT_NAME} ${areg_SRC} ${mcrouter_SRC})
+add_executable(${PROJECT_NAME} ${mcrouter_SRC})
```
biwiki: Would be nice to replace ${PROJECT_NAME} with mcrouter${AREG_BIN_EXT}

Ali-Nasrolahi: Sure :)

Ali-Nasrolahi:
library output defined in `conf/user.conf` which indicates:

```cmake
set(ProjBuildPath "build/${CrossCompile}${Toolset}/${OpSystem}-${Platform}-${Config}")
set(ProLibDir "${ProjOutputDir}/lib")
``` 
which I believe is same as Makefile's output, is that correct?

Though I have modified `Toolset` to have corresponded value as compiler which is little bit different from Makefiles.

Output executables changed to `mcrouter${AREG_BING}`.

biwiki:
> library output defined in `conf/user.conf` which indicates:
> 
> ```cmake
> set(ProjBuildPath "build/${CrossCompile}${Toolset}/${OpSystem}-${Platform}-${Config}")
> set(ProLibDir "${ProjOutputDir}/lib")
> ```
> 
> which I believe is same as Makefile's output, is that correct?
> 
> Though I have modified `Toolset` to have corresponded value as compiler which is little bit different from Makefiles.
> 
> Output executables changed to `mcrouter${AREG_BING}`.

By using Makefile, the shared library path is areg-sdk/product/build/g++/linux-x86_64-release/bin/libareg.so
The shared library should be placed in 'bin' if I could remember correctly. For Windows, it is expected that the executable and library are in the same directory ...
And should mcrouter be placed in 'bin' as well?

Ali-Nasrolahi:
I assumed based on `/conf/makefile/user.md` is should be like that.

```makefile
# The project static library directory
ProjLibDir      := $(ProjOutputDir)/lib
# The project binary output directory
ProjBinDir      := $(ProjOutputDir)/bin
```

Nevertheless, I've changed it to `bin`, however for executable output couple issues are in the way.
So I suggest we move on for now and after refining configs structure we get back to it.
> Specifically, due to some issues with commands that must be defined before some else, keeping current Makefile structure is tricky.

aregtech:
@Ali-Nasrolahi ,

Jus want to inform the logic of path structure:

- all outputs should be in the folder `product`. This folder right now contains only 1 subfolder `build`. Later, it will contain subfolder `gen` (or `generated`) for all generated files. Probably `proj` for all project files (done for IDE)
- The `build` subfolder contains other subfolder(s) with information about HW architect, compiler, OS and the build information (release / debug). Right now I don't remember by heart the sequence. The best would be to compile with `make` and see the outputs.
- Finally, all object files are output in `obj` subfolder. All static libraries are output in `lib` subfolder. In Windows, when compile as a shared library (DLL), it generates `areg.lib` and `areg.dll` files. The `areg.lib` is located in the `lib` subfolder. And the `bin` subfolder contains all shared libraries and executables, as well as configuration files. In the same example of Windows, the `areg.dll` goes to `bin`. Under Linux it is straight forward --  `areg.a` is in the `lib` and `areg.so` in the `bin`. 
 
The purpose is to simplify compiler / linker options:
1. add `product/gen` in the include compiler option;
2. add `xxx/lib` in the include of linker option;
3. collect all binaries / executables in the `xxx/bin`.

Could you please let me know whether your changes have these structure of file paths?

Ali-Nasrolahi:
Thank you @biwiki. That actually clarified some important spots.

Currently, Cmake output does not *fully* structured on what you have explained, but shortly will  be.

Again the reason is, I tried to replicate Makefile structure, so it'll be easier for you to maintain later on. However as I mention those issues, I had to change it a little - that is exactly why I added `pre-project.cmake`.

Please let me know if any other structure has conflict with what you have in mind so I work on them and the new structure becomes as close as possible to your standard. (Hopefully in close future PRs)

biwiki (approved):

biwiki:
Gents,
This is not a comment, but an improvement to consider.
CMake allows us to put all its outputs to a _build_ directory. Please take a look at this chapter for source vs binary directories
https://cmake.org/cmake/help/book/mastering-cmake/chapter/Getting%20Started.html#directory-structure
Having said so, all the binaries including the library should go to CMake binary directory.
Suppose that someone tries to compile areg-sdk with cmake like:
$ mkdir build
$ cd build
$ cmake ../
$ make -j 8
Then build directory will become CMake binary directory ....
The great benefit of this is to allow parallel compilation of the source base, here we can think about compiling all the projects with all different supported compilers/options in parallel on a powerful server.
autotools has similar feature like that ...
areg-sdk/product/build is a common path in the source directory, and using it for the build outputs may break parallel builds in the above scenario.
PS: I was looking for libareg.so in the CMake binary directory :cry:

aregtech:
@biwiki ,

i didn't get really what is the exact problem. Maybe I'm missing something.

- `areg-sdk/product/build` is not a common path for the sources, it is common path for the binaries.
- The common path for the sources are `are-sdk/gen`, which will contain generate files, split by subfolders. For example, the file located in `my-project/some/service/MyService.siml` will generate files in the folder `areg-sdk/product/gen/my-project/some/service/...`
- Depending on compiler, HW, OS and build option we have path separation like `areg-sdk/product/build/linux-x86-gcc-debug/xxx` or `areg-sdk/product/build/win32-x86-vc142-debug/xxx`, where each compiles makes own outputs in the `xxx`

Please correct me if I understood wrong your question.

aregtech:
@biwiki and @Ali-Nasrolahi,

Gents, what do you think to have regular video call? 
Let's organize regular community meetings. 
There are some projects that use the engine, maybe they would be also interested to join the community.

Ali-Nasrolahi:
@biwiki and @areg,

By removing/commenting this line from `framework/CMakelists.txt` **libareg** will be included in binary directory.
```cmake
set_property(TARGET areg PROPERTY LIBRARY_OUTPUT_DIRECTORY ${ProBinDir})
```
I must discuss the fact that we do not need to restructure the outputs directoriey if the only feature your looking for is multi-threaded compiling.

I've tried this method in *areg root* directory.

```bash
$ cmake CMakeLists.txt -B build/ -D CMAKE_RUNTIME_OUTPUT_DIRECTORY="../product/build"
$ cd build
$ make -j 8
```
By using `CMAKE_RUNTIME_OUTPUT_DIRECTORY`, we specify binray output directory - in our case `mcrouter.out`.

It works nicely.

Although reconsidering the structure would not be necessarily an unpleasant idea.(i.e. maybe some features you needed that wouldn't be convenient to implement with bare Makefiles.)

In any case, I'm open to any idea or reconsideration.

aregtech (approved):
Thank you!

aregtech:
@Ali-Nasrolahi and @biwiki 
Guys, let's have a video call and discuss these things there. If needed, we can share the screen for better understanding.

Ali-Nasrolahi:
I have no problem with video call, but due to time differences I assume my free period will be imconvient for you.
Though if you believe my presence is beneficial and necessary , then we'll do something about it.

It's your decision @aregtech.

biwiki:
Yeah, this should be up to @aregtech to consider and decide :smiley: 

And it is about parallel builds, not the multi-thread build with `-j`

Let's assume that an user decides to make use of the library and build their continuous integration pipelines like:
```text
Build    #1                     #2                             ...
         <setup binary dir>     <setup binary dir>
         make -j 8              make areg=shared -j 8
         make test              make test
         <packaging>            <packaging>
         make clean             make clean
```
As the current design, both build `#1` and `#2` share the same output directory under areg-sdk/product, they can both write to the same file at the same time?
And what would the user do at `make clean`, if they clean up everything at areg-sdk/product then that would cause a big issue since files are being used by other pipelines.
`make clean` is supposed to remove every output from the build, and if they have to remove individual files then they have to know exactly which files are generated from the build.
By using _modern_ build systems, normally people keep all their outputs in the binary directly as stated in the document.

Ali-Nasrolahi:
Sorry, I must have understood your comment in wrong way. I agree that's different. 

P. S. guys could add me to your discord server so we could continue this discussion there.

aregtech:
Currently, `make clean` removes the complete `product/build` folder. True that it will cause problem and true that it should only remove the "own" binaries. I would appreciate very much if you fix this issue. But also not a big problem right now.

About video call / regular meeting, we are already a community. If we want to established and grow, we should have regular meetings for various topics. I'll move this discussion offline.

PR #53, opened 2022-05-13
CMake integration

Ali-Nasrolahi:
Good evening @aregtech and @biwiki.

I'd like to present  new progress on CMake integration to `areg-sdk`.

Summery of new features and improvements:

- Comments added to few instructions for clarification.
- Some configuration settings have modified for better syntactical improvement.

- Examples (01 to 09) is now compiled and linked by CMake.
  - examples  `10_locsvc` through `17_winchat` (plus `00_helloservice`) has different structure than others, therefore they need different method other than what had been implemented right now in `examples/CMakeLists.txt`.
    > one of such difference is having multiple executables in a single example.

Next PR will include rest of example support.

To compile and test the output, I  used following commands at root of `areg-sdk` :

```bash
$ cmake CMakeLists.txt -Bbuild

$ cd build
$ make -j 4
```

As our previous discussion, right now all examples output is **still** included in `build` directory along with `mcrouter.out`.
Which after decision  been made on  output directory of executables, we could modify CMake configuration as needed easily.

As always, open to any changes / modification. :)


examples/CMakeLists.txt
```diff
@@ -1,10 +1,46 @@
...
+	08_service
+	09_svcmulti
+)
+
+# Iterating over fileList and extract each source file,
+# to create separate target and executable.
+foreach(item IN LISTS fileList)
+	unset(SRC)
+	set(itemDir ${AREG_EXAMPLES}/${item})
+	set(myDir ${itemDir}/src)
```
biwiki: Suggesting to rename myDir to (sub)ProjDir.

examples/CMakeLists.txt
```diff
@@ -1,10 +1,46 @@
...
+		target_compile_options(${item} PRIVATE ${CompileOptStatic})
+	endif()
+
+	# -l options for compiler
+	target_link_libraries(${item} areg m  stdc++ rt pthread ${LDFLAGS})
+	set_target_properties(${item}
+		PROPERTIES
+		CXX_STANDARD 17
+		CXX_STANDARD_REQUIRED ON
+		)
```
biwiki: Would it be possible to move/define these to common cmake file?

examples/09_svcmulti/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+include_directories(${AREG_EXAMPLES}/09_service/)
```
biwiki: Sorry but I am not sure what this is for. Could you please help explain it a bit?

examples/09_svcmulti/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
...
+include(${myDir}/CMakeLists.txt)
```
biwiki: Perhaps try starting using file glob :-)
Hopefully, something like this can work for us
```
file(GLOB files "${myDir}/*")
list(APPEND SRC ${files})
```

biwiki (commented):
It is nice to see now we can compile examples as new sub-projects :-)
Would you please help check a few cosmetic comments from me?
Thanks.

Ali-Nasrolahi:
Absolutely :)

examples/CMakeLists.txt
Ali-Nasrolahi: `myDir` renamed to `subProjDir`.

examples/09_svcmulti/CMakeLists.txt
Ali-Nasrolahi: At first, even for including core files, using `file (GLOB ..)` method was one of the easy method to just add everything together and indeed it did work. Although this method was discouraged by many, because even though it is an easy approach to add CMake to project, it lacks many benefit of an explicit source list.

Based on same rule, I tried for an explicit approach. 

I must say, there's no absolutely **any force** for using specific method. If you believe using `file(GLOB)` has more benefits for our situation, implementing such idea  wouldn't be  a problem. :)

examples/09_svcmulti/CMakeLists.txt
Ali-Nasrolahi: Well, in `main.cpp` file for including header files, this method is used:

```c
#include "src/ServicingComponent.hpp"
```

which is not a path from `areg-sdk` root directory, therefore by using `include_directories()`, we gives CMake valid path to them. In this case `src/ServicingComponent.hpp`.

Ali-Nasrolahi:
I resolved (answered) requested modifications.
Please let me know if I misunderstood the problem or question.

Thanks

biwiki (approved):
Thanks! LGTM now :-)

examples/09_svcmulti/CMakeLists.txt
biwiki: Then should it be`${AREG_EXAMPLES}/09_svcmulti/` instead?

Ali-Nasrolahi: Oh yes, sorry typo :)

aregtech (approved):
Thank you very much, @Ali-Nasrolahi

PR #65, opened 2022-11-11
Readme has been updated.

Ali-Nasrolahi:
Build section had minor changes to be more compatible for CMake build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ project(${PROJECT_NAME} VERSION ${PROJECT_VERSION} LANGUAGES C CXX)
 
 include(${CMAKE_CONFIG_DIR}/common.cmake)
 
-option(BUILD_TESTS "Build examples" ON)
+option(BUILD_TESTS "Build examples")
 
 include_directories(framework/)
 include(${AREG_BASE}/CMakeLists.txt)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,11 +17,12 @@ project(${PROJECT_NAME} VERSION ${PROJECT_VERSION} LANGUAGES C CXX)
 
 include(${CMAKE_CONFIG_DIR}/common.cmake)
 
-option(BUILD_TESTS "Build examples")
+option(BUILD_TESTS "Build unit tests")
+option(BUILD_EXAMPLES "Build examples")
 
 include_directories(framework/)
 include(${AREG_BASE}/CMakeLists.txt)
 
-if(BUILD_TESTS)
+if(BUILD_EXAMPLES)
     include(${AREG_EXAMPLES}/CMakeLists.txt)
 endif()
--- a/README.md
+++ b/README.md
@@ -98,10 +98,20 @@ An example to get source codes and compile under **Linux**. You'd need at least 
 $ git clone https://github.com/aregtech/areg-sdk.git
 $ cd areg-sdk
 ```
+
 ```bash
-# Step 2: Compile sources from terminal by calling: make [all] [framework] [examples]
-$ make all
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -98,10 +98,20 @@ An example to get source codes and compile under **Linux**. You'd need at least
...
+
 ```bash
-# Step 2: Compile sources from terminal by calling: make [all] [framework] [examples]
-$ make all
+# Step 2: Create a subdirectory for CMake cache files and change directory to it.
+$ mkdir build && cd build
+
+# Step 3: Initialize CMake cache and build systems configuration.
+# Enable examples compilation by using BUILD_TESTS flag.
+$ cmake .. -DBUILD_TESTS=ON
```
aregtech: - Can we rename `BUILD_EXAMPLES`? Because there will be unit tests and this flag we can use for that
- By default, we can have `build all` (or `BUILD_ALL`, this should be set without option)
- `build framework` -- build only framework.
- `build examples` -- build framework + examples
- `build tests` -- build framework + unit tests (planned to integrate in next version)

aregtech (changes requested):

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ project(${PROJECT_NAME} VERSION ${PROJECT_VERSION} LANGUAGES C CXX)
 
 include(${CMAKE_CONFIG_DIR}/common.cmake)
 
-option(BUILD_TESTS "Build examples" ON)
+option(BUILD_TESTS "Build examples")
```
aregtech: Let's reserve `BUILD_TESTS` for unit tests and `BUILD_EXAMPLES` for examples.

README.md
```diff
@@ -130,7 +140,7 @@ _Compile AREG SDK sources and examples:_
 |**Linux**| Open gnome-terminal in Linux and call “_make_” to compile with POSIX API.|
 
 > 💡 Compilation with _Eclipse_ under Windows might require to switch the Toolchain. For example, `Cygwin GCC`.<br>
-> 💡 For Linux the default compiler is `g++`. Set preferred C++17 compiler in [conf/make/user.mk](./conf/make/user.mk) file.
+> 💡 For Linux the default compiler is `g++`. Set preferred C++17 compiler in [conf/cmake/pre-project.cmake](./conf/cmake/pre-project.cmake) file.
```
aregtech: Can we please move all configuration specific settings in `user.cmake`?
As well, the `user.mk` is relevant for the `make`. Let's keep this. Instead, we can add separate sub-chapters (headlines) building with different tools. You can add texts and i can format it. Anyway there are other documents to update.

Ali-Nasrolahi: Unfortunately @aregtech , this set of configuration which right now only consist of setting  compiler directory; **must** be included *before* calling to `project` and by putting it inside of `user.cmak` or `common.cmak` it'll break multiple configuration which needs to be included *after* `project`.

As for the change, I'll revert it then for now.

CMakeLists.txt
Ali-Nasrolahi: I added `BUILD_EXAMPLES` and kept `BUILD_TESTS` for later use.

aregtech (approved):
@Ali-Nasrolahi , thank you!

README.md
aregtech: @Ali-Nasrolahi , OK. Understood. But let's also think how to find a solution. Or at least whether it makes sense to think about solution :)

aregtech:
Approving these changes.
The `readme.md` needs some changes, but I'll do it later.

Ali-Nasrolahi:
No problem @aregtech :)

PR #82, opened 2022-12-05
Copy init files and binary suffix in linux

### LINKED ISSUE #74: Copy config files after compilation
**Motivation**
1. After compilation, the developer should be able to start any project without additional effort. At this moment, to start IPC project or to configure logging, after compilation with `cmake`, the developer needs manually create `config` subdirectory and copy `log.init` and `router.init` files in the folder of binaries. For example, compiling with `msvc` and `make` automatically do this action.
2. `make` under `Linux` compiles and creates binaries with extension `.out` and `msvc` compiles binaries with extension `.exe`. 
3. Both, `make` and `msvc` compile with incremental build, which is much faster and does not require to recompile all projects.

**What to do**
1. After compilation of binaries with `cmake`, create `config` subfolder in the binary output folder `bin` (i.e. in `bin/config`) if it is not existing, and copy files [log.init](https://github.com/aregtech/areg-sdk/blob/master/framework/areg/resources/log.init) used by logging and [router.init](https://github.com/aregtech/areg-sdk/blob/master/framework/areg/resources/router.init) used for IPC.
2. When compile under Linux, add `.out` extension to executable binaries, i.e. the `mcrouter` is `mcrouter.out`. No need to make this change for Windows, because the compiler automatically adds extension `.exe`.
3. Support incremental build with `cmake` at least for `Debug` configuration, so that there will be no need to rebuild all projects after each small change.

Ali-Nasrolahi:
Fixes #74.


aregtech:
@Ali-Nasrolahi , thank you very much for the changes. Can you please merge changes in the `main` branch. This `20220701-candidate` branch is deprecated and I'll delete soon.

aregtech:
@Ali-Nasrolahi,
I tried to merge into the `master`, but it as well wants to update the `readme.md` file. I think, it  make sense to delete this PR and create new one without `readme.md`

Ali-Nasrolahi:
@aregtech,
Sorry I thought candidate branch will stay that's why I did not made the changes directly on master :)
I remove this PR and move the changes to `master` then.

PR #93, opened 2022-12-14
feat: basic GTest integration

harshit496:


framework/CMakeLists.txt
```diff
@@ -38,3 +38,18 @@ set_target_properties(${EXE}
     CXX_STANDARD ${aregCXXStandard}
     CXX_STANDARD_REQUIRED ON
     )
+
+# GTest setup
+enable_testing()
```
aregtech: @harshit496 , thank you for changes. It looks better some notes:

1. This code should not be in the `areg-sdk/framework/CMakeLists.txt`, but in the [areg-sdk/CMakeLists.txt](https://github.com/aregtech/areg-sdk/blob/master/CMakeLists.txt) file, which should include new `tests/CMakeLists.txt` file.
2. Please as well use `BUILD_TESTS` in [CMakeLists.txt](https://github.com/aregtech/areg-sdk/blob/master/CMakeLists.txt#L19) to enable compiling tests. Something like this (prototype):
```
 if (BUILD_TESTS)

   set(GOOGLE_TEST_BASE "${AREG_BASE}/thirdparty/googletest")
   set(TESTS_BASE "${AREG_BASE}/tests")
   set(UNIT_TEST_BASE "${TESTS_BASE}/units")
   include(${TESTS_BASE}/CMakeLists.txt)
...
 ````
3. Please do not forget to check whether the codes / folder of `${GOOGLE_TEST_BASE}` exists (for example, check existence `thirdparty/googletest/CMakeLists.txt`). If not, output error message with the hint to clone `googletest` module and break compilation of unit tests.
4. All test relevant actions / setups should be done in the file `tests/CMakeLists.txt`
5. All unit test relevant actions should be done in the file `tests/units/CMakeLists.txt`
6. Please keep in mind that the folder `units` may have subfolders. For example, `units/string` to make string relevant unit tests, or `units/common` to make common setups, etc. This is optional, but may happen to avoid having huge list of sources in one directory.

The rest looks good

Comment (harshit496) [reply framework/CMakeLists.txt]: @aregtech @Ali-Nasrolahi need your suggestion here. 
to address the comments above I modified the `areg-sdk/CMakeLists.txt` to have the following
```
if(BUILD_TESTS)
    set(GOOGLE_TEST_BASE "${AREG_SDK_ROOT}/thirdparty/googletest")
    if(EXISTS "${GOOGLE_TEST_BASE}/CMakeLists.txt")
        # For Windows: Prevent overriding the parent project's compiler/linker settings
        include("${GOOGLE_TEST_BASE}/CMakeLists.txt")
        set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
        set(TESTS_BASE "${AREG_SDK_ROOT}/tests")
        set(UNIT_TEST_BASE "${TESTS_BASE}/units")
        include(${TESTS_BASE}/CMakeLists.txt)
    else()
        message(SEND_ERROR "Please clone the submodules using command\n
        'git clone --recurse-submodules https://github.com/aregtech/areg-sdk.git'")
    endif()
endif()
```

although the googletest CMakeLists file (`areg-sdk/thirdparty/googletest/CMakeLists.txt`)
```
if(BUILD_GMOCK)
  add_subdirectory( googlemock )
else()
  add_subdirectory( googletest )
endif()
```
 doesn't have that relative path when CMake would execute the line `include("${GOOGLE_TEST_BASE}/CMakeLists.txt")` and hence ends up with the error

```
CMake Error at thirdparty/googletest/CMakeLists.txt:36 (add_subdirectory):
  add_subdirectory given source "googlemock" which is not an existing
  directory.
```

what is the best way to handle this?

Comment (aregtech) [reply framework/CMakeLists.txt]: @Ali-Nasrolahi , can you please have a look? I'll also try to have a look today late evening.

File: CMakeLists.txt [hunk 2]
```diff
@@ -16,7 +16,7 @@ project(${PROJECT_NAME} VERSION ${PROJECT_VERSION} LANGUAGES C CXX)
 
 include(${CMAKE_CONFIG_DIR}/common.cmake)
 
-option(BUILD_TESTS "Build unit tests")
+option(BUILD_TESTS "Build unit tests" OFF)
 option(BUILD_EXAMPLES "Build examples" ON)
 
 include_directories(framework/)
```
aregtech, line 22: I would also include `thirdparty`
```cmake
include_directories(framework/)
```
but can do that later. It's optional.

aregtech (approved):
Looks good. Many thanks @harshit496 !

@Ali-Nasrolahi , if you don't have any notes, I'll merge the changes.

harshit496:
@aregtech dont merge this yet because this wont compile when BUILD TESTS is on. I just pushed the code so that @Ali-Nasrolahi can refer to the code to help fix the issue mentioned above.

aregtech:
@harshit496, ah, I see. I thought you fixed :) 
OK, then please let me know.
I still didn't have chance to have a look.

Ali-Nasrolahi:
@harshit496 could you post the cmake's output when you trying to compile?

harshit496:
@Ali-Nasrolahi here you go.
```
@harshit496 ➜ /workspaces/areg-sdk/build (60-integrate-gtest) $ cmake ..
-- >>> 'Release' build for 'Linux' platform with compiler '/usr/bin/g++', ID 'GNU'
-- >>> Build output folder '/workspaces/areg-sdk/product/build/gnu/linux-x86_64-release'
CMake Error at thirdparty/googletest/CMakeLists.txt:36 (add_subdirectory):
  add_subdirectory given source "googlemock" which is not an existing
  directory.
Call Stack (most recent call first):
  CMakeLists.txt:33 (include)


-- Configuring incomplete, errors occurred!
See also "/workspaces/areg-sdk/build/CMakeFiles/CMakeOutput.log".
See also "/workspaces/areg-sdk/build/CMakeFiles/CMakeError.log".
```

aregtech:
It looks like because CMake does not switch the directory (working directory) and it remains in `AREG_BASE`
I made a small change like this and the error has gone:
```cmake
if(BUILD_GMOCK)
  add_subdirectory( ${GOOGLE_TEST_BASE}/googlemock )
else()
  add_subdirectory( ${GOOGLE_TEST_BASE}/googletest )
endif()
```

harshit496:
I had initially tried that too but the cmake change you made is for the third party googletest cmakelist file which would be essentially recursively cloned so this change wont be there and it will fail again right?

aregtech:
@harshit496, another important thing, can you please bring these changes not to the `master` branch, but to the `feature/60-integrate-google-test` branch. All GTest related things we should do in that branch and when the branch compiles, either you or I can create a PR to merge changes with `master`.
You can as well delete this PR and create another to merge with the branch.

harshit496:
@aregtech sounds good. will do

aregtech:
We should search in [discussions](https://github.com/google/googletest/discussions) whether someone had similar problem. I do believe this is not the first project :) 
Another work around could be changing working directory and calling `cmake` executable process to build googletest.


P.S. Thank you switching the branch :)

harshit496:
@aregtech @Ali-Nasrolahi I've finally fixed the issue. GoogleTest official docs suggest using the `add_subdirectory()` command and after adding that it finally works!

Please take a look. Thanks!

Ali-Nasrolahi:
LGTM, nice work :)

aregtech (approved):

aregtech:
@harshit496, thank you very much.
And please don't forget to star the repository :)

I'm currently testing. Looks like there are some paths are missing. Probably I can fix.
When all builds run, I'll merge with master.

PR #124, opened 2023-02-15
Feature/122 unit tests for strings

harshit496:


tests/units/StringTest.cpp
```diff
@@ -3,7 +3,7 @@
 
 TEST(StringTest, makeInt32)
```
aregtech: - Please add comment for the test.
- Please add `Test` prefix to the method.

tests/units/StringTest.cpp
```diff
@@ -3,7 +3,7 @@
...
 {
-    char *test = "52";
+    const char *test = "52";
     int32_t result_int32 = 52;
```
aregtech: For both cases please use `constexpr`

tests/units/StringTest.cpp
```diff
@@ -3,7 +3,7 @@
...
     EXPECT_EQ(String::makeInt32(test), result_int32);
     EXPECT_EQ(typeid(String::makeInt32(test)), typeid(result_int32));
```
aregtech: No need to test `typeid`.
Avoid using `typeid`, `dynamic_cast` and `exception` whenever it is possible. Only if there is a real need.

tests/units/StringTest.cpp
```diff
@@ -38,4 +38,16 @@ TEST(StringTest, getSubstring)
     
     result = String::getSubstring(next, " ", &next);
     EXPECT_EQ(result, "0123");
+}
+
+TEST(StringTest, makeUInt32)
+{
+    const char *test = "52";
+    uint32_t result_uint32 = 52;
```
aregtech: For both variables, please use `consexpr`

tests/units/StringTest.cpp
```diff
@@ -38,4 +38,16 @@ TEST(StringTest, getSubstring)
     
     result = String::getSubstring(next, " ", &next);
     EXPECT_EQ(result, "0123");
+}
+
+TEST(StringTest, makeUInt32)
```
aregtech: - Please comment the test.
- Please add a `Test` prefix to the method.

tests/units/StringTest.cpp
```diff
@@ -38,4 +38,16 @@ TEST(StringTest, getSubstring)
...
+{
+    const char *test = "52";
+    uint32_t result_uint32 = 52;
+    EXPECT_EQ(String::makeUInt32(test), result_uint32);
+    EXPECT_EQ(typeid(String::makeUInt32(test)), typeid(result_uint32));
+    
+    uint16_t result_uint16 = 52;
+    EXPECT_EQ(String::makeUInt32(test), result_uint16);
+    EXPECT_NE(typeid(String::makeUInt32(test)), typeid(result_uint16));
```
aregtech: No need for `typeid`

aregtech (changes requested):
@harshit496 ,
thank you for changes. It is a good start.
Please let's keep same / similar syntax, be closer to C++17 and avoid testings, which give nothing.
I avoid using `typeid`, `dynamic_cast` and `try-catch` blocks, because they increase the size of binaries. As well as if a returns type is `int`, there is no need to check the type.
Please do not hesitate to discuss if you other opinion. I'm open to listen and get better ideas.

tests/units/StringTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+    int32_t result_int32 = 52;
+    EXPECT_EQ(String::makeInt32(test), result_int32);
+    EXPECT_EQ(typeid(String::makeInt32(test)), typeid(result_int32));
+    
+    int16_t result_int16 = 52;
+    EXPECT_EQ(String::makeInt32(test), result_int16);
+    EXPECT_NE(typeid(String::makeInt32(test)), typeid(result_int16));
+}
+
+TEST(StringTest, getSubstring)
```
aregtech: Please comment the test and add `Test` prefix.

aregtech:
Hello @harshit496 
do you still have interest to write unit tests?

harshit496:
Hi @aregtech sorry I was traveling so couldn't get to it. I can pick it up now. Thanks for the reminder and sorry for delaying this.

aregtech (approved):
Thank you, @harshit496 
It is a good start.

harshit496:
Thanks! Although @aregtech I actually had some more tests that weren't committed. Nvm I will get them in another MR then

PR #186, opened 2023-07-13
183 log message reader thread

### LINKED ISSUE #183: Log message reader thread
**Description**
In the current implementation, the host of logs is connecting as a client to the remote lob service to send logs. To have communication with the service, there is a need to implement the reader thread, so that the application can receive instructions from the service to change the log priority.

**Remark:**
This is issue uses protocols created in #177.

aregtech:
Implemented the reader thread to receive messages from remote logging service.
Fixed an important bug in the map templates. It had wrong variable names.


aregtech (commented):
reviewed one more time

PR #207, opened 2023-07-23
202 connection lost notification

### LINKED ISSUE #202: Connection lost notification
**Description:**
At the moment the service consumer receives a connection notification with the 2 states: connected and disconnected, where _disconnect_ state is considered for any state where the communication with service provider is not possible, including the cases where there is no network connection (connection is lost). It is useful instead of boolean value `true` and `false` to deliver the disconnect reason. 

Examples of use:
1. a Service Provider completes the job and shuts down --> the service consumer should receive disconnect message to reset the states. Because on next start the Service Provider will have another state.
2. a router is down, so that Service Provider and Service Consumer loose connection, but they would like to continue with the current state. --> when router connection is lost, provider and consumer receive message "connection is lost" and save the state. When connection is established, they continue communication.

aregtech:
Implement and fixed connection lost for service providers and service consumer objects.


aregtech (commented):
checked documentation. ignored checking examples.

PR #209, opened 2023-07-26
206 refactor the service manager

### LINKED ISSUE #206: Refactor the ServiceManager
**Description:**
The service manager is one the most important components that manages the connection between service providers and service consumers. By time it has been grown and the code became quite complicated. Considering the fact that there are still multiple features are in the pipeline, this object will continue to grown.

**Solution**
The code should be simplified and easy to read. For this reason, while the code did not became more complex, there is a need to refactor the object, make it simpler and easier to understand.

aregtech:
Refactored the service manager to make it easier to track and read.


aregtech (commented):
checked comments

PR #218, opened 2023-08-07
2017 Fixed local datetime conversion

### LINKED ISSUE #217: Fix local DateTime for Linux (high)
**Describe the bug**
There are 2 ways to convert to local times. The unit test to [compare time](https://github.com/aregtech/areg-sdk/blob/54f7e54b10696bdc9668305a8aae40c14160bdbc/tests/units/DateTimeTest.cpp#L231) fails, reporting 2 converted local times are not equal.

**To Reproduce**
Uncomment [this line](https://github.com/aregtech/areg-sdk/blob/54f7e54b10696bdc9668305a8aae40c14160bdbc/tests/units/DateTimeTest.cpp#L231) in the [DateTimeTest.cpp](https://github.com/aregtech/areg-sdk/blob/54f7e54b10696bdc9668305a8aae40c14160bdbc/tests/units/DateTimeTest.cpp) compile under Linux and run.

**Expected behavior**
Indifferent how the UTC time is converted to the local time, it should show same time.

- OS:  Linux
- Hardware: amd64
- AREG SDK version [e.g. 1.5.0]

aregtech:
Fixed local date-time conversion for the Linux. There was a difference when converting 
`<gmt in seconds> --> <struct tm> -> <time_t> -> <struct tm>`

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -2,8 +2,9 @@ name: CMake
 
 on:
   push:         # Keep empty to run on each branch when push the code. Otherwise use branches: [ master ]
-  # pull_request: # Set to master to run only when merge with master branch
-  # branches: [ master ]
+    branches: [master]
+  pull_request: # Set to master to run only when merge with master branch
+    branches: [ master ]
 
 env:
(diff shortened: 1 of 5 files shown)


vukihub (approved):
looks good

PR #225, opened 2023-08-17
224 option parser

### LINKED ISSUE #224: Console option parser
**Description:**

Since the amount of tools are raising and the majority of them will have console version, there is a need to implement a console option parser generic object that parses the command line and creates the option list.

aregtech:
Implemented the option parser as part of extensions. Added unit tests to test the options. Found and fixed a bug in the TEString class.


aregtech (commented):
checked.

vukihub (approved):
looks good.

PR #227, opened 2023-08-18
219 Implemented start and quit logger.

### LINKED ISSUE #219: Start and stop of logging service
**Description:**
The logging service is a separate process, which can run either as a system service or console application. There should be possibility to start and stop service via service manager, and stop / quit on command line, similar to `mcrouter`.

aregtech:
Implemented start and quit logger.


vukihub (approved):
looks good.

PR #231, opened 2023-08-23
Integrated logger send and receive threads

### LINKED ISSUE #220: Network connection threads of logging service
**Description:**
The logging service should have separate threads to accept connections and receive data, and another thread to send data. The threads should automatically start when the logger starts and shutdown then application is closing.

vukihub:
Integrated receive and send threads in the logger. Updated logger.init file, fixed the build scripts


aregtech (approved):
looks good.

PR #232, opened 2023-08-25
220 network connection threads of logger

aregtech:
- Implemented console commands for `mcrouter` and `logger`.
- Checked the connection reconnect while process is running.
- Found and fixed bug in the TEMap.
- Found and fixed but in the Thread.


vukihub (approved):
lg

PR #235, opened 2023-08-26
Fixed router pause and restart issue.

### LINKED ISSUE #233: After router pause and restart the example 13 stops IPC
**Describe:**
The example 13 stops IPC after router pause and restart.
The applications are connected, each sends request to register services, but it does not start communication. It seems that after reconnect, the system stuck on service provider / consumer registration.

**Hint:**
One of possible solutions could be cleaning list of providers and consumers on the `mcrouter`, because after pause, there is anyway no connection and the lists must be cleaned.

aregtech:
Fixed the issue of pausing and restarting message router.
Fixed issue of reconnect.
Fixed issue of preparing thread to receive events.
Fixed bug in the example.

### DIFF
--- a/examples/06_threads/src/main.cpp
+++ b/examples/06_threads/src/main.cpp
@@ -100,8 +100,9 @@ protected:
 /************************************************************************/
 
     /**
-     * \brief   Triggered before dispatcher starts to dispatch events and when event dispatching just finished.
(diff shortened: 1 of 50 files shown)


vukihub (approved):
looks good

vukihub (approved):
checked fix

PR #241, opened 2023-08-27
233 after router pause and restart the example 13 stops ipc

### LINKED ISSUE #237: Restructure thread closing
**Description:**
The `Thread` object has 3 ways to complete:
1. `destroyThread` -- which may as well terminate the thread if timeout expires;
2. `shutdownThread` -- same as destroy, but with infinite timeout. As well it may differ in each class implementation;
3. `completionWait` -- waits the thread to complete the job without terminating it.

These might be a little confusing for developers and in most cases it nearly does same thing. For this reason, these 3 methods should be merged in one function, for example, all in `shutdownThread`. As well, all object that call these methods should be reviewed and probably a little bit restructured.

### LINKED ISSUE #238: Applications complete the job with delay
**Describe:**
Sometimes it may take a time until the applications complete the job and exit. It is better visible when application runs with multiple threads and components. For instance, in the example 13, noticed that it may take seconds until application exists. It may take 10-30 seconds.

**Action:**
Investigate to figure out the reason of such delay.
Make refactoring if needed and fix the issue. The application should start and stop as fast as possible. The main reason of delay is because of waiting the threads to complete the job. So that, the exit procedure should be optimized.

aregtech:
1. Improved multithreading communication;
2. Fixed a bug creating POSIX timer;
3. Fixed a bug adding existing listener in the StubBase

### DIFF
--- a/docs/HelloService.md
+++ b/docs/HelloService.md
@@ -44,7 +44,7 @@ When creating a service provider and service consumer, it is important to keep i
 5. Instances of the service provider and its consumer can be in the same thread.
 6. The _Role Names_ of the _Local_ services are unique within the same process, and the _Role Names_ of the _Public_ services are unique within the same network.
 7. Service consumers know the availability of the service via the notification callback `void serviceConnected(NEService::eNetConnection status, ProxyBase& proxy)`, where `status` parameter indicates the connection status.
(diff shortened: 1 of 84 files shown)


vukihub (approved):
looks good

PR #244, opened 2023-09-01
Added new example 23 to check mixed of communications

### LINKED ISSUE #239: Consumers are not notified on first subscription
**Description:**
Observed in the codes that Service Consumers (Proxy) are not immediately notified on data update when first time subscribe on the data (attribute). The Service Provider (Stub) adds the consumer in the list and does not send an update.

**Expected behavior:**
If the data state is valid, the Service Consumers (proxies) must immediately receive the data update notification when first time subscribe on data.

**Solution:**
1. There can be 2 different consumers that subscribe: local and public;
2. If the subscription is new and data state is valid, they must immediately receive the update message and the message must be targeted. Otherwise, all subscribed consumers without any reason receive update notification. So that, the notification must have precise target.
3. If Provider received instruction to subscribe on data and it is presented, then:
   - Ignore sending update if it local or public in the same process (proxy source is local).
   - Send update it is remote public, because the misbehavior can be due to lost connection.
   - Ignore sending update for local services if they are already present.

aregtech:
There were several fixes in previous commits. In order to check the communication and fixes, created 2 new examples to demonstrate the data update notifications. All checks prove that communication work fine. The minor issue, which is neither a bug, nor features:
- When the service consumer looses network connection, the data (attribute) of the remote service is not automatically marked as invalid. One side, this may keep the information of the last available status. On the hand, since it is unknown how long the connection is lost, it might make sense to invalidate the data on Consumer side.

In addition, made changes in `Makefile`. Make it more clear.


vukihub (approved):
Looks good. The documents should be updated.

vukihub (approved):

vukihub (approved):

PR #246, opened 2023-09-04
Fixed connection issue and notification assignment

### LINKED ISSUE #240: The public service consumer must remain connected if network connection lost
**Description:**
Observed that if a Service Provider (Stub) has connected Consumers (Proxies), which one is running on local process and the other is on remote, and if connection is lost, then Service Consumer, which is running on the same process, may also stop receiving messages.

This should be investigated, observed and tested well. Probably with additional example with simple features.

In addition, check that Service Provider removes all notification subscription of the disconnected Service Consumer, so that:
- On the disconnect all requests, notifications and responses are cleaned, because makes no sense to process;
- On the re-connect can restart everything to pass the current state of data;
- If a request is pending, it should be canceled.

### LINKED ISSUE #245: Wrong messages on data subscription
**Description:**
While was testing _example 24_, found 2 bugs:
- When assign on data update, the client does not immediately receive update notification. If data is not available yet, it should notify with data state "not available", then receives the actual.
- After data subscription, if a component has multiple instances of subscribers to the same publisher data, it receives more events than necessary.

**Expected behavior**
- After subscription, the client should immediately receive the actual data and the data state flag indicating the validity of the data.
- If a component has a more than one subscribers that are subscribe on the same data of the same publisher should not increase (double or triple, etc.).

aregtech:
* Fixed the issue with connections (bug #240)
* Made sure that subscribers receive the exact expected numbers of events (bug #245)
* Added new example to check the bug #245 
* Update example of unblocking to check the number of received responses.


vukihub (approved):
looks good to me.

PR #249, opened 2023-09-18
Generic implementation of service connection

### LINKED ISSUE #248: Generic solution of connectivity
**Motivation:**
The _Message Router_ and the _Logger_ have many similarities regarding communication and messaging. There should be a generic solution where both service should share same logic to connect, reconnect, disconnect, send and receive messages, and dispatch the messages.

**Description:**
The current implementation of the AREG framework is focused on the connectivity with one instance of the _Message Router_. The implementation of the connection to the any other service requires many customization, which blows up the code, makes the code less readable and in case of bugs, makes it more difficult to fix. Since the connectivity has many similarities, it makes sense to find common generic solution, which can be used in multiple cases. This would also be a first step to implement multiple communication channel, as it is described in the issue #31.

aregtech:
Implemented the generic solution for the connectivity to use to connect with message router and log collector.
Changed the sequence number to the 64-bit unsigned integer.
Changed the code generator to match the 64-bit sequence number.


vukihub (approved):
looks ok

PR #251, opened 2023-09-18
use 64-bit digits by reference instead of values

### LINKED ISSUE #250: Reference to the 64-bit integer
**Description:**
In the recent changes the sequence number is switched from 32-bit into the 64-bit unsigned integer. Since the areg engine is designed to run 32-bit system as well, this conversion may cause addition CPU operation when copy the value in the parameter or on return. To avoid this, use references not only for the sequence number, but all 64-bit integers.

aregtech:
Converted 64-bit digits to be passed or returned by references, instead of value. In 64-bit environment this may play no role, but in 32-bit environment we may escape additional operations.


vukihub (approved):
looks good

PR #253, opened 2023-10-07
Implemented log message dispatcher on client side.

### LINKED ISSUE #236: Logger client message dispatcher
**Describe:**
Any client application to the log collector should have a logging message dispatcher to re-translate, dispatch and execute messages.  `areg` framework should have integrated message dispatcher, which is activated when the logging is enabled, and should dispatch. If a binary is created without logs, the dispatcher can be excluded from the build.

aregtech:
Implemented log message dispatcher on client side to respond on queries to send the list of scopes, update the scopes and save log configuration.


vukihub (approved):
looks good.

PR #262, opened 2023-11-19
Feature/256 update configurator

### LINKED ISSUE #255: Optimize configuration file
**Description:**

Currently, there are 3 configuration files: _router.init_, _log.init_, _logger.init_. All files use same format. By time, when the list of helper application increase, the number of init files may as well increase. So that, it is important to use universal configuration file and format to save all configuration in a single file, where each application may extract own portion of information. In addition, there are several objects that read the configuration file. For example, the router connection and log setup are done by different objects. To optimize the solution, there should be:

- The configuration should be setup centralized, where each object that needs an information can query and get own list of configuration data.
- The centralized configuration should be able to save data in the file.

### LINKED ISSUE #259: Integrate TraceController with new congigurator
**Describe:**

In the existing implementation of logging configuration, when saving the scope priorities in the `log.init` file, the TraceController has a function that makes the saving. As AREG framework new implementation will have centralized configuration (Configurator), the TraceController should be modified to set information of scopes in the Configurator then make saving into the file without harming or changing any other settings in `.init` file.

### LINKED ISSUE #260: Optimize all configurations
**Describe:**
Because of centralized configuration, minimize all configuration objects, remove all not used methods and leave only methods to query property and value. For more details see #256

### LINKED ISSUE #261: Temporary Property
**Description:**
There should be 2 types of properties -- readonly and writable, where writable property may overwrite the settings of readonly. Since application global settings can be stored in the configuration, the developer may need to store some settings, which are temporary and should not be saved. For this reason, the writable properties should have additional _temporary_ flag, indicating that the property change is temporary and should not be saved in the configuration file when the developer requests to save. So that, the saving ignores such properties and they are not changed.

aregtech:
- Refactored configuration. Now all configurations are in the `areg.init` file
- Implemented dynamic configuration, so that any data can be stored in `.init` file
- Implemented _global_ for all modules configuration and _module specific_ (process specific) configurations
- Fixed multiple bugs in templates and objects
- Fixed example 17
- Implemented delay of service connected event
- Implemented possibility to read configuration from multiple `.init` files to set configuration manual


aregtech (commented):
checked

PR #266, opened 2023-11-22
264 - fixed pause and restart of mcrouter under Linux.

### LINKED ISSUE #264: mcrouter does not pause propertly
**Describe:**

When pausing the `mcrouter` it stucks and reacts on no other commands. The only way around is to kill the application and start again.
Happens always.

aregtech:
Fixed the bug when mcrouter was stuck during pausing.
The bug was happening under Linux compile with enabled extensions.


aregtech (commented):
checked changed.

PR #267, opened 2023-11-24
265 wrong formatting to output message

### LINKED ISSUE #265: Wrong formatting to output message
**Describe:**

When print command `-h` to print the help, it outputs messages not formatted, which is difficult to read.

**Tested app:**

Compiled with GCC.
Extensions are enabled.
Tested under Linux (Ubuntu).

aregtech:
Fixed printing help for mcrouter and logger.


aregtech (commented):
rechecked.

PR #270, opened 2023-12-01
Fixed resizing ring stack, segmentation fault and exit logging threads

### LINKED ISSUE #269: NetLog message queue initialization
**Describe the bug**
When logging scopes are initialized, it is instantiating TraceManager and the NetLog object, which queue size is initialized with default value and before the application configuration is set. This may cause several problems:
1. The queue size is initialized by using default value and any change in the configuration file is ignored.
2. The queue is initialized indifferent whether the network logging is enabled or not.
3. Since log messages can be queued only if network logging is enabled, it makes no sense to reserve space, because all log messages will be dropped.

**To Reproduce**
Change value "log::*::remote::queue" and debug by setting break-point.

**Expected behavior**
Do not reserve space before configuration.
There can be 2 solutions:
1. Read configuration before application initialization, so that the queue can be properly sized (reserve).
2. Ignore initialization of queue in the constructor, so that it can be initialized later. All generated messages before initialization will be dropped.

aregtech:
Made multiple fixes of bugs:
1. Fixed logic of reading configuration and resizing ring stack of the messages in the NetLogger;
2. Fixed segmentation fault issue happening after reserving space in the ring stack.
3. Fixed issue of exiting read and receive threads when run NetLogger. After closing the threads where not ending.
4. Fixed issue of delaying exit threads.
5. Made changes in the layouts to be able to log cookies and module name.

### DIFF
--- a/framework/areg/trace/private/NetTcpLogger.cpp
+++ b/framework/areg/trace/private/NetTcpLogger.cpp
@@ -42,7 +42,7 @@ NetTcpLogger::NetTcpLogger(LogConfiguration & logConfig, ScopeController & scope
 
     , mScopeController  ( scopeController )
     , mIsEnabled        ( false )
-    , mRingStack        ( logConfig.getStackSize(), NECommon::eRingOverlap::ShiftOnOverlap )
+    , mRingStack        ( 0, NECommon::eRingOverlap::ShiftOnOverlap )
 {
 }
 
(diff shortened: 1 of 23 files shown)


framework/areg/trace/private/NetTcpLogger.cpp
```diff
@@ -57,18 +57,18 @@ bool NetTcpLogger::openLogger(void)
...
             mIsEnabled = true;
-            resizeStack(mLogConfiguration.getStackSize());
             applyServiceConnectionData(host, port);
             mLock.unlock();
             result = connectServiceHost();
         }
         else
         {
-            resizeStack(0);
+            mRingStack.reserve(0);
```
vukihub: Would this release the ring stack?

framework/areg/trace/private/TraceManager.cpp
```diff
@@ -276,10 +276,6 @@ void TraceManager::readyForEvents( bool isReady )
     {
         DispatcherThread::readyForEvents( false );
         TraceEvent::removeListener( static_cast<IETraceEventConsumer &>(self( )), static_cast<DispatcherThread &>(self( )) );
-
-        mLoggerFile.closeLogger( );
-        mLoggerDebug.closeLogger( );
-        mLoggerTcp.closeLogger( );
```
vukihub, line 282: If sure that there is no need to close loggers here, please add assertions.

vukihub (changes requested):
Some slight changes are requested.

vukihub (approved):
Looks good.

PR #276, opened 2023-12-06
FIxed mcrouter reacting on commands

### LINKED ISSUE #275: Cannot exit mcrouter
**Describe the bug**
`mcrouter` does not react on any command.

**Expected:**
Should react on commands.

aregtech:
Fixed `mcrouter` to react on commands
A preparation from other ticket is present.


vukihub (approved):
ideally would be having fix only from mentioned ticket.

PR #277, opened 2023-12-08
Tested and fixed query scopes, fixed reserve method of shared buffer.

### LINKED ISSUE #26: Implement logging service
**Goal**
Currently AREG SDK already has powerful logging / tracing mechanism and currently it makes logs only in the file. The logging is configurable and the configuration is defined in the `log.init` file. The logging supports _log scopes_ and _priority_, which can be dynamically changed during runtime. But at the moment there is no mechanism to send message to set scopes.

Similar to `mcrouter`, there should be a separate process (`logging service`) to run on any machine that is able to collect the logs and communicate with the applications via messages (predefined protocol). The logging service should also forward logs to the log viewer (planned to develop).

In order to make these achievements there is a need to define communication messages (communication protocol) between the service and applications, which is able:
- receive process / application name
- request the list of scopes, activation status and log priority
- forward logs to the log viewer
- save logs in a readable format in the file.
- send commands to the application such as:
    * change active status and log priority of the scope;
    * change logging configuration (including service IP / port nr.)
    * save current configuration in the config file (`log.init`).

Every application should be able to save configuration in the config file, so that can use on next start.
- Send logs to log viewer (planned to develop)

**Motivation**
Logging is a powerful feature and tool. Sometimes debugging an application to analyze bug behavior is not always effective. For example, if a bug is related with the timing, it would be very difficult and sometimes impossible to find the bug. In such situations the logging is irreplaceable and very helpful tool.

**Background information**
AREG SDK already has logging (see module `trace`), which makes many jobs such as 
* dynamic change of priority
* read, parse and save logging config in `log.init` file.
The missing part is the logging service and bidirectional communication with the service. The implementation of the `mcrouter` can be a good example of developing service that uses network communication.

### LINKED ISSUE #274: Query log scopes
**Describe:**
Implement query log scopes, so that the log collector or log observer can query applications and receive list of scopes.

aregtech:
- Tested and fixed query log scopes;
- Fixed bug when reserve space in shared buffer -- move pointer and then reserve space;
- Optimized creating remote message;
- Optimized forwarding log remote messages.


framework/areg/trace/private/NETrace.cpp
```diff
@@ -30,32 +30,7 @@
...
-                , NEService::SEQUENCE_NUMBER_NOTIFY             // rbhSequenceNr
-            }
-            , { static_cast<char>(0) }
-        };
-
-        return _messageRegisterScopes;
-    }
-
-    const NEMemory::sRemoteMessage& _getLogUpdateScopes(void)
+    const NEMemory::sRemoteMessage& _getLogEmptyMessage(void)
```
vukihub, line 39: Comment the methods.

vukihub (changes requested):
A local function is without comment. Please describe the purpose of the method.

vukihub (approved):
looks good

PR #278, opened 2023-12-10
Fixed bug in example and in inputting string from console

### LINKED ISSUE #272: Wrong alignment on console when run example 14
**Describe the bug**
When run example 14 it has wrong text alignment: new line of text starts in the new line, but in the next column where the previous text ended. It does not look nice.

**To Reproduce**
1. Compile project under _Linux_;
2. Enable AREG extension and enable compiling examples;
3. Start `mcrouter` for IPC communication;
4. Start client and service parts of example 14.

*Result:* on client side it has wrong text alignment.

**Expected behavior**
The text should be nicely aligned. There should be no difference whether it runs with extensions or not.

- OS: Ubuntu Linux (WSL)
- Compiler: clang++14
- AREG SDK version: latest

aregtech:
1. Fixed the bug in the example 14. For any reason, the C++ call `std::cout` did not output the string on console properly. Replaced that with the `printf()` and added `\n\r` at the end of every output message to align properly.
2. Improved method `Console::_osWaitInputString` by removing `\n` at the end of the string left by `gets_s() / fgets()` methods. Otherwise, the comparison of strings didn't work properly, because  for example `"ew" != "ew\n"`, where `"ew\n"` is a string returned by `gets_s() \ fgets()`.

### DIFF
--- a/examples/14_pubtraffic/pubclient/src/TrafficLightClient.cpp
+++ b/examples/14_pubtraffic/pubclient/src/TrafficLightClient.cpp
@@ -72,17 +72,17 @@ inline void TrafficLightClient::outputState(NESimpleTrafficLight::eTrafficLight 
     switch (lightState)
     {
     case NESimpleTrafficLight::eTrafficLight::LightRed:
-        std::cout << "Light: RED ..." << std::endl;
+        printf("Light: RED ...\r\n");
         break;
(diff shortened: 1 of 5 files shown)


vukihub (approved):
Looks fine

PR #279, opened 2023-12-10
Delete events in sorted event stack.

### LINKED ISSUE #191: Removing events in the sorted stack (low prio)
**Describe the bug**
Application may crash when exit.
The SortedEventStack deletes events in destructor. It may cause of crash if the event object does not exist. For example, if a static Exit event is deleted before SortedEventStack of any other thread (for example, ServiceManager thread).

**To Reproduce**
I [removed the deleting of events](https://github.com/aregtech/areg-sdk/pull/190/files#r1265006221) from destructor.
To reproduce, bring the changes back and it may randomly happen.

**Expected behavior**
No crash.

- OS: Windows
- Compiler: MSVC
- AREG SDK version. Before [this](https://github.com/aregtech/areg-sdk/pull/190/files#r1265006221) change
 
**Additional context**
When application exits, no thread or thread object should exit.
Make changes to dump the thread information to make sure that all threads exit OK.
The behavior observed when run example 14.

aregtech:
Brought back the calls to delete events in the destructor of the sorted stack and made tests. The original bug seems is fixed.

### DIFF
--- a/framework/CMakeLists.txt
+++ b/framework/CMakeLists.txt
@@ -1,6 +1,6 @@
 
 # ###########################################################################
-# User sepcific settings
+# User specific settings
 # ###########################################################################
 include("${AREG_BASE}/areg/CMakeLists.txt")
 include("${AREG_BASE}/extend/CMakeLists.txt")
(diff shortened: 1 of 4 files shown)


vukihub (approved):
looks fine.

PR #282, opened 2023-12-17
Fixed bug: when logger crashes, the client should reconnect.

### LINKED ISSUE #281: Clients reconnect to the logger.
**Describe:**
When the applications are connected to the `logger` and the `logger` is crashing or disconnecting, on restart, the applications do not connect anymore.

**Expectation:**
The clients should always connect to the `logger` whenever it is available.

aregtech:
When logger crashes, the client socket connection did not request to re-connect, so that the client application should be restarted as well. Fixed the issue by triggering 'connection lost' event when failed to receive or send data.


vukihub (approved):
Looks fine

PR #287, opened 2023-12-28
The logger service client object.

### LINKED ISSUE #286: Log observer client object
**Description:**
 Since the log observer is receiving logs from the logger service, it should have a client socket object to connect to the logger to start receiving messages. So that there is a need:
- Instantiate a socket client connection object that connects to the logger;
- Have send and receive threads;
- Send and receive basic handshake message;

aregtech:
The client object to send and receive message from the logger service. It has the basic implementation and ready to send receive message, as well as to dispatch the messages.


vukihub (approved):
Looks good.

PR #290, opened 2024-01-01
Fixed logger service messages

### LINKED ISSUE #283: Prepare observer API library
**Description:**
Since AEREG framework has a fixed communication messages for the logs, including log observers, there is a need to provide a library (shared or static library) that contains API with methods that can be used in any observer application. In the future this library will extend the API.
- Check the API and messages
- Check the sequencies
- Check the applications

aregtech:
- Checked the logger service API
- Fixed messages
- Added log of instance connection.


vukihub (approved):
looks good

PR #300, opened 2024-01-09
Fixed reconnect to the logger.

### LINKED ISSUE #296: Logobserver does not reconnect to the log service
**Describe:**
When the log service (`logger`) is shutting down and restarting again, the connected log observer console application (`logobserver`) does not try to reconnect again. The solution is to stop logging (option `-s`) and starting again (optiont `-r`). Unlike to log observer, the client applications immediately connect.

**Expected behavior:**
The log observer should try to reconnect and connect to the log service `logger` as soon as it starts, so that the clients can log messages.

aregtech:
- On disconnect and connection failure, trigger 'connection lost';
- Fixed an assertion when the state is still connected, but the cookie is invalid;
- Reverted by default build with extended console feature.


vukihub (approved):
Looks OK

PR #308, opened 2024-01-17
307 integrate sqlite embedded database

### LINKED ISSUE #307: Integrate SQLite embedded database
**Describe:**
Currently the logs are saved in the file as a plane text. In order to create and use powerful log viewer, it makes sense to save logs in a structure that is easy to read and write, and apply filters. This could be a embedded database like SQLite, which can be embedded in any application, which is able to run SQL scripts and which is portable.
The advantages of using embedded database like SQLite:
- embedded, can run in any application;
- easy to integrate;
- supports basic SQL scripts and allows to use the power of SQL;
- allows to structure and create relational tables;
- allows filters and views;
- portable

**Note:**
At this very moment still OK to discuss and offer alternative solution. Requirement:
- The logs may reach a few GB. 
- Should be fast and portable, should not require any additional software installation like database server.
- Only one instance can write, but multiple instances can read.

vukihub:
- Integrated the SQLite as a static library;
- Created Makefile script to build googletest with `make` tool.


aregtech (approved):
looks fine. thank you.

PR #311, opened 2024-01-18
310 fix codeql analytics warning

### LINKED ISSUE #310: Fix CodeQL analytics warning
**Describe:**
CodeQL reported security warning because of SQLite3 C-code sources. Tried to exclude the complete third-party codes from analytics, but faced the problems that workflow action was failing. To solve this:
- fix the CodeQL scripts that it does not fail anymore;
- fix SQLite3 codes that there is no security warning.

aregtech:
- Fixed SQLite3 codes to fix warnings.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,13 +13,12 @@ name: "CodeQL"
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


vukihub (approved):
ok

PR #316, opened 2024-01-23
Implemented saving logs in the SQLIte database.

### LINKED ISSUE #314: Log messages in the SQLite database
**Describe:**
After providing the minimum SQLite3 wrapper objects, on log observer side insert log messages into the database.
- Inserting the logs should be a part of the 'log observer API` library, so that it can save logs in the DB
- On every log start, create new database with the mask ./log/logs_%time%.db3
- Create 3 tables: `instances`, `scopes` and `logs`;
- On every connection of an instance, insert new entry in the `instance` table;
- On every scope registration or update, insert scopes in the the `scopes` table;
- Every new log should be inserted into the `logs` table.
- The observer (API) does not query, `update` or `delete` an entry.

**Additional information:** At this moment, the log observer API will provide the minimum functionalities to log message into the database. In the future, there will be new API available to split or merge databases, to query logs, etc.

aregtech:
- Implemented saving logs in the SQLite database.
- Extended the log observer API and integrated SQLite database logging in the library.
- Extended the log observer console application to save the logs in SQLite database.


framework/extend/db/private/LogSqliteDatabase.cpp
```diff
@@ -0,0 +1,466 @@
...
+#include "areg/base/File.hpp"
+#include "areg/base/NEMath.hpp"
+#include "areg/base/Process.hpp"
+#include "areg/base/Thread.hpp"
+#include "areg/component/NEService.hpp"
+#include "areg/trace/NETrace.hpp"
+
+namespace
+{
+    constexpr std::string_view  _sqlCreateTbVersion
```
vukihub, line 35: Would be nice to add couple of comments to explain each SQL script.

vukihub (changes requested):
Please add documentation to the SQL scripts. The rest looks fine.

vukihub (approved):
looks fine.

PR #318, opened 2024-01-24
Fixed: create log database when compile project without logs.

### LINKED ISSUE #317: Test log observer
**Describe:**
Since the log observer is implemented and it contains multiple features, it is important to test with various builds:
- Build log observer API as shared and static library, make sure that log observer console application compiles and runs well.
- Build AREG library without logs (AREG_LOGS = OFF), make sure that it still receives and saves log messages.
- Some logs in the database indicate no thread name, find the reason why it is so and make sure that the behavior is connect. _Hint:_ this may happen if the logs are generated in the main thread of the application. In this case it might have no name.

aregtech:


vukihub (approved):
looks good.

PR #321, opened 2024-01-27
Fixed inserting instances after logger reconnect.

### LINKED ISSUE #319: Failed create new entry on reconnect to logger
**Describe:**
1. Start lob observer console application;
2. Type `-r` to trigger connect to `logger` service. **Result:** this creates log database (_OK_)
3. Start `logger` log collector service;
4. Start `mcrouter` and/or any other application that generates logs. **Result:** this creates entries in the `instances` table of the log database, as well scopes and logs the messages (_OK_)
5. In the command line of `logger` service, type `-q` to quit application. **Result:** This updates the entries of `instances` table and makes instances as disconnected. As well marks the scopes as inactive (_OK_)
6. Restart the `logger` service:
    **Results:**
    - After connecting to the `logger` service, the log observer receives list of connected instances. Since `logger` makes new start, it generates cookie IDs starting from 256. Then the log observer tries to create new instance in the `instances` table, it fails, because there are already entries with the same cookie ID exist. (_NOK_)
    - It creates list of active scopes of the applications (_OK_);
    - Logs message in database (_OK_).

**Expected behavior:** 
On reconnect to the `logger` service, it should make new entries of the instances.

aregtech:
- Fixed SQL scripts to update instances after `logger` reconnect;
- Slight changes in constants and slight cleanups.
- Enabled AREG extended library features by default when compile with MSVC.

### DIFF
--- a/framework/areg/component/NEService.hpp
+++ b/framework/areg/component/NEService.hpp
@@ -222,9 +222,9 @@ namespace NEService
     typedef enum class E_DisconnectReason : uint16_t
     {
           ReasonUndefined               = 0     //!< Undefined disconnect reason.
-        , ReasonRouterDisconnected      = 1     //!< The router is disconnected.
-        , ReasonRouterLost              = 2     //!< The router connection is lost.
-        , ReasonRouterRejected          = 4     //!< The router rejected connection.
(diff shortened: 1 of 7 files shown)


vukihub (approved):
looks good.

PR #325, opened 2024-02-03
323 datetime mismatch in windows and linux

### LINKED ISSUE #323: DateTime mismatch in Windows and Linux
**Describe:**
The DateTime mismatch if run applications under different OS like Windows and Ubuntu. The time generated in Linux indicate 1672, where it cannot be less than the year 1900.

**To Reproduce**
1. Start `logger` and `mcrouter` as a services in the Linux on local machine (for example, in WSL);
2. Start any client-service application (for instance, the _example 16_) in the Linux (WSL)
3. Start the log observer console application in Windows.

**Result:** Shortly, get assertion in the log observer, which is indicating that the DateTime 64-bit numeric value points year 1672, which is absolute incorrect.

**Expected behavior**
Indifferent in which environment the applications run, since the system indicates correct day and time, there should not be any problem with date and time conversion neither in Windows nor in Linux.

**Additional context**
Try to clone sources in WSL or in the Virtual Machine, compile and run application to make sure that the subsystem has right day and time settings. Otherwise, the problem can be in the subsystem settings.

aregtech:
Fixed but in date-time utilities. Made sure that the time is in UTC, Unix epoch, and converted them

### DIFF
--- a/framework/areg/base/private/NEUtilities.cpp
+++ b/framework/areg/base/private/NEUtilities.cpp
@@ -39,28 +39,20 @@ namespace NEUtilities
 
     extern void _osSystemTimeNow( NEUtilities::sSystemTime & OUT sysTime, bool localTime );
 
-    extern void _osSystemTimeNow( NEUtilities::sFileTime & OUT fileTime, bool localTime );
-
-    extern TIME64 _osSystemTimeNow( void );
-
     extern TIME64 _osConvToTime( const NEUtilities::sSystemTime & IN sysTime );
 
     extern void _osConvToSystemTime( const TIME64 & IN timeValue, NEUtilities::sSystemTime & OUT sysTime );
     
-    extern void _osConvToSystemTime( const NEUtilities::sFileTime & IN fileTime, NEUtilities::sSystemTime & OUT sysTime );
-
-    extern void _osConvToFileTime( const NEUtilities::sSystemTime & IN sysTime, NEUtilities::sFileTime & OUT fileTime );
-    
-    NEMath::eCompare compareLargeIntegers( const NEMath::uLargeInteger & lsh, const NEMath::uLargeInteger & rhs )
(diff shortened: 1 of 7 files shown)


framework/areg/base/private/posix/NEUtilitiesPosix.cpp
```diff
@@ -21,79 +21,13 @@
...
-    {
-        time_t secs = static_cast<time_t>(quad / SEC_TO_MICROSECS);
-        time_t rest = static_cast<time_t>(quad % SEC_TO_MICROSECS);
-        outSecs     = secs;
-        outMilli    = static_cast<unsigned short>(rest / MILLISEC_TO_MICROSECS);
-        outMicro    = static_cast<unsigned short>(rest % MILLISEC_TO_MICROSECS);
-    }
-
-
     uint64_t _osGetTickCount( void )
```
vukihub, line 31: Please add commnet either here of in NEUtilities.cpp file where the OS specific methods are declared as `extern`.

framework/areg/base/private/NEUtilities.cpp
```diff
@@ -20,62 +20,78 @@
...
 // static local functions
 /************************************************************************/
 namespace NEUtilities 
 {
-    extern const char * _osGenerateName(const char * prefix, char * out_buffer, int length, const char * specChar);
-
     extern uint64_t _osGetTickCount( void );
 
-    extern bool _osConvToLocalTime( const sSystemTime & IN utcTime, sSystemTime & OUT localTime );
+    extern void _osSystemTimeNow( sSystemTime & OUT sysTime, bool localTime );
```
vukihub: Please document the OS specific methods.

vukihub (changes requested):
Please document the methods.

vukihub (approved):
looks fine.

PR #330, opened 2024-02-10
Single unit test executable.

### LINKED ISSUE #329: Merge unit test executables
**Describe:**
In the current implementation, the unit tests are compiled in the following way:
1. In MSVC they are all compiled as `areg-unit-test` executables;
2. In `cmake` they are compiled as separate executables, where  each unit test file is compiled as a separated executable.
3. In `make` they are not compiled at all.

**Needs:**
There is a need:
1. In MSVC nothing should be changed;
2. In `cmake` they all should be linked as one executable. Otherwise, when implementing more unit tests, the list of compiled executables will be increased, which is not really optimal solution;
3. In `make`, if continue using this tool, the unit tests should be as well compiled with `make` in one unit test executable.

aregtech:
Instead of multiple unit test executables, now create one executable that includes multiple tests. Same as it is done in MSVC.

### DIFF
--- a/conf/cmake/functions.cmake
+++ b/conf/cmake/functions.cmake
@@ -450,3 +450,46 @@ macro(macro_declare_project project_name project_alias)
     macro_declare_project_ex(${project_name} ${project_alias} ${project_alias})
 endmacro(macro_declare_project)
 
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


vukihub (approved):
looks good.

PR #332, opened 2024-02-11
194 stop using submodules

### LINKED ISSUE #194: stop using submodules
**Description**
In the current implementation of areg-sdk, it uses `gtest` as a submodule. This requires additional git command to download sources of linked submodules. In addition, submodule force other project to stay with specific version of the dependent project. The cloning procedure and the the dependencies are not optimized.

**Solution*
The possible solution is to to simplify the cloning and dependencies as much as possible, considering following problems:
1. Compilation with CMake
    - With CMake it is possible to check whether the `gtest` package is present and if necessary, download.
    - Check whether it is enough to download once of this process is repeated each time.
2. Compilation with MSVC:
    - Investigate whether it is enough to use the `gtest` module of the MSVC
    - Investigate how it is possible to download `gtest` if it is missing
3. Compilation with Make
    - In the current implement there is support of `gtest` in the Make. But should investigate how to implement with the logic close to CMake.

If there is any problem that at least (1) and (2) cannot be solved, we stay with the submodule solution.

aregtech:
- Stopped using submodules.
- Use fetching instead submodules.
- Removed `Makefile", disable build with `make` tool since supporting is getting more complicated.
- Added `googletest` package in MSVC
- Set fetching modules, packages and builds in `./product` directory.


vukihub (approved):
Checked the most critical parts.

PR #333, opened 2024-02-14
175 cmake changes to generate code

### LINKED ISSUE #175: CMake changes to generate code
**Description:**
The projects, which use Service Interface document file (files with extension `.siml`) contain generated codes, which are manually integrated in the projects by CMakeLists.txt files. Since these files are generated and the sources are kind of fixed, it makes sense to modify the CMakeLists.txt file to generate and compile files during build.

**Required:**
1. Modify the CMake to generate and build static libraries from Service Interface document files. Something like `addServiceInterface(myproject, myproject/HelloWordl.siml)`, where the `myproject` is the name of the project and static library and `myproject/HelloWordl.siml` is the relative path to the Service Interface document.
2. There should be possibility to link the static library with the target binary, something like `linkProject(target, myproject)`.
3. Modify the CMake build to add an option that will only generate files and create MSVC project files.

**Note:**
The description might be not really clear. Please start the discussion here and ask questions to @aregtech before starting modifying.

aregtech:
Use code generator during configuration, instead of predefined generated sources. Compiled code generator to use in lower versions. The lowest version is 1.8


vukihub (approved):
.

PR #334, opened 2024-02-16
Extended DateTime.

### LINKED ISSUE #271: Extend DateTime class
**Description**

The DateTime contains a simple methods to get time, but it is still missing such methods like duration, conversion and simple arithmetic operations. it should be fully compatible with `std::chrono`  or should be at least easily convertible.
In addition, it as well requires mode unit tests.

aregtech:
Extended DateTime, made slight cleanups.


vukihub (approved):
looks ok.

PR #337, opened 2024-02-21
213 reduce the size of areg library

### LINKED ISSUE #213: Reduce the size of areg library
**Description**
Since depending on compilation flags some features might be disabled, it makes sense to include sources to the binary only if they are needed. This might reduce the size of binary. For example, if we compile sources without logs, it totally makes sense to exclude sources in `trace` subfolder of `areg`.

**Solution:**
Either separate the `trace` subfolder of `areg` in a separate project to be linked with `areg` and others, or exclude them from the build.
Consider the cross dependencies between submodules of `areg` (for example, `trace` uses classes from `base` and `components` use features from `trace` (make logs of areg). It means that the separation may take longer time rather than excluding sources from the build.

aregtech:
Reduced the size of `areg` library when compile without logs. Fixed unit tests running logging checkups when logs are not enabled.


vukihub (commented):
looks fine

vukihub (approved):

PR #340, opened 2024-03-05
338 service interface xsd

### LINKED ISSUE #338: Service Interface XSD
**Describe:**

Based on [Sample.siml](https://github.com/aregtech/areg-sdk/blob/master/docs/Sample.siml) file, create an XSD that is matching the requirement of `.siml` structure.

aregtech:
Integrated XSD, checked with `Sample.siml`


aregtech (commented):
rechecked

PR #344, opened 2024-06-22
343 Fixed warnings and the bug related with missed switch-case

### LINKED ISSUE #343: Fix warnings during compilation
2 warnings are reported when compile with GCC:
1. In `SystemServiceBase.cpp` file it seems the `case` statement is missed;
2. When compile the SQLite static library, get warning `cc1: warning: command-line option ‘-std=c++17’ is valid for C++/ObjC++ but not for C`

aregtech:
- Fixed the missed `case` in switch-case;
- Fixed warnings with the missed keyword `override`;
- Fixed `cmake` script to avoid a warning (as well linker error) when compiling sqlite3 library.


vukihub (approved):
looks fine

PR #345, opened 2024-06-27
TEArrayList unit test setup

### LINKED ISSUE #196: Unit test for TEArrayList
**Description:**
[TEArrayList](https://github.com/aregtech/areg-sdk/blob/master/framework/areg/base/TEArrayList.hpp) is a wrapper of `std::vector` adapted to areg-sdk needs. It is needed to add unit tests to test all public methods and to cover as much codes as possible.

**Solution:**
1. There is a [./tests/units/](https://github.com/aregtech/areg-sdk/tree/master/tests/units) folder in the project to add unit tests.
2. Add in that folder a new unit testing file, for example `TEArrayListTests.cpp`, and write unit tests.
    - The unit tests should run at least with `int`, `unsigned int`, `float` / `double` and `std::string` and `String` / `WideString` types. 
    - The tests with the custom types, like custom structures, are welcome, but it is optional.
    - If the amount of testing methods are to large, they can be split and can be implemented either in a separate folder (for example `array`) or can be split to separate files like `TEArraListStringTest.cpp`.
    - Each created file should be added in the [CMakeLists.txt](https://github.com/aregtech/areg-sdk/blob/master/tests/units/CMakeLists.txt) and in the [areg-tests](https://github.com/aregtech/areg-sdk/blob/master/tests/areg-unit-tests.vcxproj) MSVC project.
    - Each `.cpp` file should have a list of tested methods at the top
    - Each test method should have convenient name that is describing the purpose of the test, and should have short description.
3. The tests should run. In case if a bug found, create a ticket and mention the priority in the title, like `[bug]`. Each created ticket should have at least 2 labels: `bug` and `unit test`.

aregtech:
Adding unit tests.


vukihub (approved):
looks fine.

vukihub (approved):
looks good

PR #346, opened 2024-07-13
unit tests of copy and move methods of TEArraylist

aregtech:
Added new unit test to check `copy` and `move` methods of TEArraList object.


vukihub (approved):
good

PR #348, opened 2024-07-19
unit test to test remove methods of TEArrayList

aregtech:
unit test to test remove methods of TEArrayList


vukihub (approved):
looks fine. Commenting tests would be better :)

PR #349, opened 2024-07-21
unit test for sift method of TEArrayList

aregtech:
unit test for sift method of TEArrayList.
Fixed bugs


vukihub (approved):
looks good. thank you.

PR #350, opened 2024-07-23
196 unit test for tearraylist

aregtech:
Next things were updated:
* added tests to test TEArrayList entries;
* added tests to test streaming;
*  update test of constructor
* updated TEArrayList and fixed syntax error in template.


vukihub (approved):
Looks fine!

PR #373, opened 2024-09-08
bugfix/372 cmake make fails after a clean

### LINKED ISSUE #372: Cmake - make fails after a clean
**Describe the bug**
Using cmake, make command fails after clean previous build with `make clean`

**To Reproduce**
- `cmake -B build/`
- `cd build`
-  `make clean`
- `make -j 8`

**Expected behavior**
Build should succeed

**Screenshots**
NA

- OS: Ubuntu
- Version 22.04
- Hardware x86
- Compiler [g++
- AREG SDK version: master
 
**Additional context**
```
/usr/bin/ld: cannot open output file ../product/build/gnu-c++/linux-64-x86_64-release-shared/bin/libareg.so: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/areg.dir/build.make:3057: ../product/build/gnu-c++/linux-64-x86_64-release-shared/bin/libareg.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:289: CMakeFiles/areg.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```
This is probably because the `bin` directory has been removed with `make clean`

aregtech:
- created a target named `areg-dummy`, which contains pre-build commands to create directories;
- in the `functions.cmake` file in the functions `addExecutable`, `addStaticLibrary`, `addSharedLibrary` added dependency from `areg-dummy`, that the project `areg-dummy` is started and executed first;
- made `areg` library dependent on `areg-dummy`.

These all makes sure that the build of `areg-dummy` starts first and the directory structure is created before any binary is built.

Tested on WSL with Ubuntu by calling `make clean` and `make -j 8`. When call `make clean`, the output directory is deleted, so that all files with logs and testing files are deleted. And when call `make -j 8`5 it re-creates the directory structure.

### DIFF
--- a/conf/cmake/common.cmake
+++ b/conf/cmake/common.cmake
@@ -182,8 +182,11 @@ message(STATUS ">>> Java ${Java_VERSION_STRING} at location ${Java_JAVA_EXECUTAB
 message(STATUS "-------------------- CMakeLists Status Report End ----------------------")
 message(STATUS CMAKE_SOURCE_DIR = ${CMAKE_SOURCE_DIR})
 
-add_custom_target(dummy ALL COMMAND ${CMAKE_COMMAND} VERBATIM)
-add_custom_command( TARGET dummy PRE_BUILD
+# add a custom command to create output build folders if they are not existing.
+# create a dummy project to add as a dependency in all other projects.
+# this ensures that the commands to create directories are called first.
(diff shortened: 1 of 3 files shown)


biwiki (approved):
LGTM! Thanks for fixing the fault.

PR #375, opened 2024-09-23
374 fix demo project build

### LINKED ISSUE #374: Fix demo project build
**Describe the bug**
Cannot compile AREG SDK demo project when try to fetch the AREG SDK sources in the project.
It either fails complaining that `areg-dummy` project is already defined or fails when using AREG project specific variables in the CMakeLists.txt file.

**To Reproduce**
in `areg-sdk-demo` project instead of using submodules, fetch areg-sdk sources, make changes in CMakeLists.txt and include areg-sdk sources in the build. During `cmake` configuration it fails with multiple errors.

**Expected behavior**
`areg-sdk-demo` project should be possible to compile without problems.

aregtech:
Fixed the problems with CMake variables to let fetching areg-sdk sources in the external project and compile them to build applications based on areg framework.


vukihub (approved):
looks fine

vukihub (approved):
is OK

PR #380, opened 2024-09-27
optimization of headers included in the private directory.

### LINKED ISSUE #379: Hide headers in private
**Describe:**
Optimize the includes of header files used in public folder, so that mostly they are visible and accessible mostly in the source files or in headers declared in the `private` directories.

aregtech:
Optimized header files included in the public classes.


vukihub (approved):
Looks fine

PR #381, opened 2024-09-28
areg sdk installation in the cmake files #377

### LINKED ISSUE #377: Install AREG Framework binaries
**Describe:**
Make changes in the `CMakeLists.txt` file so that it could be possible to install AREG Framework binaries, header files, configuration files and tools in a separate folder. It should be possible to configure the installation destination via option passed to the `cmake` build tool.

aregtech:
Added possibility to install areg sdk binaries, includes, configuration files and tools.
After build, the user may call `cmake --install ./build` to install areg sdk package.
For more details, see description in the `user.cmake` file.


vukihub (approved):
looks fine

PR #385, opened 2024-10-01
384 Use SQLite3 installed package in the build.

### LINKED ISSUE #384: use sqlite3 package
**Describe:.**
Currently `areg-sdk` is using `sqlite3` source codes and compiles as a static library. The `vcpkg` has reference to the `sqlite3` package, which can be as well used in the `areg-sdk`.

**Actions:**
1. make appropriate changes to have possibility to use `sqlite3` package available at `vcpkg`.
2. Keep the existing source codes to use `sqlite3` either when cannot find the package or when developer forces.

aregtech:
Made changes in the cmake files to use existing SQLite3 package instead of compiling static library in the project. Up from now it is possible to use either existing in the system sqlite3 library or compile the existing in the project source.

### DIFF
--- a/conf/cmake/install.cmake
+++ b/conf/cmake/install.cmake
@@ -65,24 +65,27 @@ if (AREG_INSTALL)
     # Check and install thirdparty dependencies if needed
     if (AREG_INSTALL_DEPENDS)
 
-        # copy compiled binaries in the bin and lib directories
-        install( TARGETS sqlite3
-            ARCHIVE DESTINATION lib         COMPONENT Development   COMPONENT ${AREG_PACKAGE_NAME}
-                    PERMISSIONS OWNER_READ OWNER_WRITE               GROUP_READ WORLD_READ
-        )
+        # Check how SQLite was built
+        if (NOT AREG_FOUND_SQLITE3_PACKAGE)
+            # copy compiled binaries in the bin and lib directories
+            install( TARGETS sqlite3
+                ARCHIVE DESTINATION lib         COMPONENT Development   COMPONENT ${AREG_PACKAGE_NAME}
+                        PERMISSIONS OWNER_READ OWNER_WRITE               GROUP_READ WORLD_READ
+            )
 
-        # Copy thirdparty sources with headers
-        install(DIRECTORY thirdparty/
-                    DESTINATION thirdparty COMPONENT Development   COMPONENT ${AREG_PACKAGE_NAME}
-                            PATTERN "*.vcxproj*" EXCLUDE
-        )
+            # Copy thirdparty sources with headers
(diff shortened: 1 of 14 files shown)


conf/cmake/install.cmake
```diff
@@ -65,24 +65,27 @@ if (AREG_INSTALL)
     # Check and install thirdparty dependencies if needed
     if (AREG_INSTALL_DEPENDS)
 
-        # copy compiled binaries in the bin and lib directories
-        install( TARGETS sqlite3
-            ARCHIVE DESTINATION lib         COMPONENT Development   COMPONENT ${AREG_PACKAGE_NAME}
-                    PERMISSIONS OWNER_READ OWNER_WRITE               GROUP_READ WORLD_READ
-        )
+        # Check how SQLite was built
+        if (NOT AREG_FOUND_SQLITE3_PACKAGE)
```
vukihub, line 69: Maybe makes sence to copy dependencies?

conf/cmake/user.cmake
```diff
@@ -201,6 +203,10 @@ if (NOT DEFINED AREG_INSTALL_DEPENDS)
     option(AREG_INSTALL_DEPENDS "Disable installation of thirdparty dependencies" OFF)
 endif()
 
+if (NOT DEFINED AREG_SQLITE_PACKAGE)
```
vukihub, line 206: What about integrating "use packages" flag to use GTest package as well and all other upcoming?

thirdparty/sqlite3/Readme.md
```diff
@@ -1,3 +1,15 @@
 # SQLite sources
 
 These are _amalgamation_ sources of the SQLite database, which are taken from the [_Download_](https://sqlite.org/download.html) page of the [SQLite](https://sqlite.org/) site and not owned by Aregtech. For more information about the SQLite and licenses, please visit the official web site.
+
+## Using SQLite package
+
```
vukihub, line 35: would be nice to add more information about compiling with packages.

vukihub (changes requested):
Please see comments.

conf/cmake/install.cmake
aregtech: Can create a ticket to think about. But not with these changes.

conf/cmake/user.cmake
aregtech: Good idea. I was thinking about. I'll create a separate ticket for this.

thirdparty/sqlite3/Readme.md
aregtech: OK. Agree.

vukihub (approved):
Looks fine.

PR #392, opened 2024-10-09
391 Isolate cmake files of example

### LINKED ISSUE #391: Isolate cmake files of example
**Describe:**

Currently, the `.cmake` files of AREG framework and examples are located in the `functions.cmake`.
Some of macro and functions are not relevant to the framework and they are example specific.
These macro and functions should be isolated in a separate file and excluded from imports, so that the `functions.cmake` contain only pure framework related scripts. As well, not used macro or functions should be removed, the cmake script should be optimized.

aregtech:
- cleaned up `functions.cmake` file:
    - removed unused functions and macro
    - moved example relevant functions and macro to the appropriate `CMakeLists.txt`
    - moved unit test relevant functions to the unit test `CMakeLists.txt`
- Made changes in the `CMakeLists.txt` files
- Improved documentation and updated comments in the `functions.cmake` file.
- Optimized `CMakeLists.txt` files and structures, reduced number of files.


vukihub (approved):
Good optimization.

PR #395, opened 2024-10-12
394 macro to create binaries

### LINKED ISSUE #394: cmake macro to create binaries
**Description:**
The examples and the demo applications are using similar cmake macro to create binaries.
It would make sense to provide macro in the `functions.cmake` file to allow other applications to use.
These macro should provide possibility to build static and shared libraries, as well as executables.
If possible, as well provide possibility to set property for the resources files like `.rc` files in the Windows.

aregtech:
Defined macro to  simplify crating static and shared libraries, and executables based on AREG Framework. The macro can be used by any project.


vukihub (approved):
thank you

PR #401, opened 2024-10-19
An option to load config file from command line

### LINKED ISSUE #400: Identical command line options
**Describe:**

The `mcrouter` and the `logger` can run as OS-managed service and as console application. Make sure that both console applications have possibly same / similar options to be installed, start, runtime commands. As well, make sure that both services have possibility to load configuration file via command line option. Additionally, pay attention on `logobserver` console application that it has similar options / commands, which has `logger`.

aregtech:
Added an option to load config file from command line. The path to the config file can be relative or full path.
This give flexibility to load config file in any location, not just the hardcoded file path. 
Also, fixed bug when parsing a command line.
Fixed bug when was created `areglogger` -- the `.def` file was not properly set in MS Visual Studio.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -405,7 +405,7 @@ jobs:
         run: |
           mkdir ./${{matrix.config.name}}/
           cp -R ./${{env.BUILD_DEST}}/bin/ ./${{matrix.config.name}}/
-          cp -R  {{env.INSTALL_DEST}} ./${{matrix.config.name}}/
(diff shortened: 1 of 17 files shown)


vukihub (approved):
looks fine. There is a warning to pay attention, please fix in next commit:
`framework\aregextend/service/NESystemService.hpp(193,14): warning : enumeration value 'CMD_Custom' not explicitly handled in switch [-Wswitch-enum] [D:\a\areg-sdk\areg-sdk\product\cache\llvm-clang-cl\aregextend.vcxproj]`

PR #429, opened 2024-11-05
Feature/122 unit tests for strings

### LINKED ISSUE #122: Unit Tests for strings
**Motivation.**
The string functions are classes are essential, which are used in many cases. It is important to have unit test and very high code coverage of these objects and methods.

**Actions**
* In the [unit test](https://github.com/aregtech/areg-sdk/tree/master/tests/units) folder create at least one file to implement unit tests;
* Similar to DemoTest, include the test in the [CMakeLists.txt](https://github.com/aregtech/areg-sdk/blob/master/tests/units/CMakeLists.txt) so that it will compile separate executable.
* All test methods of unit test should be documented with short description of the test agenda.

**Additional actions**
If a bug is found (test failed), there should be a ticket created and the UT code should be included to demonstrate failure.

harshit496:


tests/units/StringTest.cpp
```diff
@@ -0,0 +1,107 @@
...
+    
+    constexpr uint16_t result_uint16 = 52;
+    EXPECT_EQ(String::makeUInt32(test), result_uint16);
+}
+
+/**
+ * Tests makeInt64() which converts a string
+ * to a int64
+*/
+TEST(StringTest, TestMakeInt64)
```
aregtech: Would be nice to add a case for negative number and positive number with the sign `+`.

tests/units/StringTest.cpp
```diff
@@ -0,0 +1,107 @@
...
+    constexpr int16_t result_int16 = 52;
+    EXPECT_EQ(String::makeInt64(test), result_int16);
+}
+
+/**
+ * Tests makeUInt64() which converts a string
+ * to a unsigned int64
+*/
+TEST(StringTest, TestMakeUInt64)
+{
```
aregtech, line 202: Would be also nice to add a case, where the `uint64_t` is greater or equal to 0x8000'0000'0000'0000 to make sure that the 64-bit conversion is correct.

tests/units/StringTest.cpp
```diff
@@ -0,0 +1,107 @@
+#include "units/GUnitTest.hpp"
+#include "areg/base/String.hpp"
+/**
+ * Tests makeInt32() func which converts a string
+ * to type int32.
+*/
+TEST(StringTest, TestMakeInt32)
+{
```
aregtech, line 39: Can you please add following cases:
- negative number conversion;
- positive number with `+` sign, like `constexpr char * test = "+123";`
-  Other non-standard cases like:
   - `constexpr char * test = "abcd";` 
   -  `constexpr char * test = "123abc";` 
   - `constexpr char * test = "abc123:'`
- Cases with various `radix`:
   ```cpp
   typedef enum class E_Radix : int8_t
   {
         RadixAutomatic    =  0    //!< Detect automatically
       , RadixBinary       =  2    //!< Binary conversion, conversion base is 2
       , RadixOctal        =  8    //!< Octal conversion, conversion base is 8
       , RadixDecimal      = 10    //!< Decimal conversion, conversion base is 10
       , RadixHexadecimal  = 16    //!< Hexadecimal conversion, conversion base is 16
   } eRadix;
   ```
- Please also check the value of the `end` parameter after calling `makeXXX`. Here is the syntax:
   ```cpp
    /**
     * \brief   Converts given string of digits to 32-bit integer
     * \param   strDigit    The string with digits. Can contain negative or positive sign in front
     * \param   radix       The base value when calculate integer.
     * \param   end [out]   If not nullptr, on output this contains value of pointer to the next character in strDigit buffer after the numerical value.
     * \return  Returns 32-bit integer
     **/
    static int32_t makeInt32( const char * strDigit, NEString::eRadix radix = NEString::eRadix::RadixDecimal, const char ** end = nullptr );
   ```

Ideally, would be nice to have parameterized tests to add various parameters and check outputs. Checkout the parameterized test example of strings in `NEStringTest.cpp`. Same is for other `makeXXX` tests.

I indeed appreciate very much your help, but then I'll have to rework and nearly overwrite everything to make tests proper :)

aregtech (changes requested):
Please read my comments.

tests/units/StringTest.cpp
harshit496: I have used the Gtest parameterize feature to parameterize the test in this test case. please take a look if that works for you and I can modify other test cases accordingly. thx

aregtech:
Thank you! Looks much better.

harshit496:
Can be merged?

aregtech:
Could you please make sure that it is compiled under Visual Studio? I tried locally and get errors.
1. The `.cpp` file is not included in the `areg-unit-tests.vcxproj` and `areg-unit-tests.vcxproj.filters` files;
2. The MSVC google test adapter does not understand `INSTANTIATE_TEST_SUITE_P` macro. In other files I use
    ```cpp
    #if defined(INSTANTIATE_TEST_SUITE_P)
    INSTANTIATE_TEST_SUITE_P()
    #else   // !defined(INSTANTIATE_TEST_SUITE_P)
    INSTANTIATE_TEST_CASE_P()
    #endif  // defined(INSTANTIATE_TEST_SUITE_P)
    ```
    An example see in file [NEStringTest.cpp](https://github.com/aregtech/areg-sdk/blob/181857e65db0e6b596fc76d0f3cbbfb2e5e7bfda/tests/units/NEStringTest.cpp#L411)

The builds in this PR passed, because new `StringTest.cpp` file is not included in MSVC build. As soon as you include, the MSBuild will fail.
@harshit496 , can you please fix the issue?

harshit496:
Ah okay. I will take a look. Thanks for pointing out

harshit496:
@aregtech I forgot to mention that I am using Linux and I was able to build it fine with cmake although it wont be able to build using msbuild right since that is windows specific?

aregtech:
@harshit496, there is GitHub action workflow to run MSBuild. If you include the file in the MSVS project, it will fail.
Another solution is, if you merge changes not with `main`, but with other branch, then I can work on that and fix. Let me know if you need a branch.

harshit496:
@aregtech I added it to MS build project. where can I find the github action workflow link to see it fail? because currently under the checks tabs on this PR it does not run any jobs. i dont know why

aregtech:
Hi @harshit496 ,

[This](https://github.com/aregtech/areg-sdk/blob/master/.github/workflows/msbuild.yml) the workflow. Your fork should also have it. As soon as you merge with your `master` branch, it will be triggered.

Alternatively, you can edit the beginning of `msbuild.yml` to compile on every push and pull request on any branch:
```yml
name: MSBuild

on:
  push:         # Keep empty to run on each branch when push the code. Otherwise, use branches: [ master ]
    # branches: [ master ]
  pull_request:
    # branches: [ master ]
```
Then you can test locally.

aregtech:
closing because it is already one year that there is no activity. there are already some string unit tests and they can be extended

PR #436, opened 2024-11-09
435 document to use areg package

### LINKED ISSUE #435: Document to use areg package
**Describe:**
Describe in details how to use `areg` package for CMake and Visual Studio, how to setup and build an application.

aregtech:
* Added detailed document to install and use `areg` package.
* made slight updates of other documents to fix links.


vukihub (approved):
looks good

PR #441, opened 2024-11-11
432 cmake configuration for cross-compiling

### LINKED ISSUE #432: CMake configuration for various platforms
**Describe:**
Currently, the build with CMake depends on the actual platform. There should be a possibility to compile CMake for different platform, like compilation for 32-bit system or compilation for ARM processors, etc.

If the task is to big, it should be split and on the first step there should be a possibility to compile for 32-bit Linux and 32-bit Windows systems. Use `AREG_BITSET` option for this task.

aregtech:
* Fixed the cross-compiling for x86, x86_64, ARM and AARCH64 CPU.
* Fixed sources to verify compilation.
* Updated documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,17 +148,14 @@ git clone https://github.com/aregtech/areg-sdk.git
 
 ### Build Instructions
 
-The **AREG SDK** is written in **C++17** and supports multiple platforms and compilers:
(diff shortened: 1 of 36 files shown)


vukihub (approved):
looks good

PR #454, opened 2024-11-28
453 file in wrong format error

### LINKED ISSUE #453: file in wrong format error
**Describe:**

When cross-compiling and if a packages  is available in the system, the compilation may fail with error message `file in wrong format`, indicating that CMake found a package, but the it is available for the other architecture. For example, if we compile for the target `x86` on the host machine `x86_64`, where the `sqlite3.so` library / package is available and it is compiled for the `x86_64` architecture, the `find_package(SQLite3)` will indicate that `found package`, but when try to link for `x86` target build, it will fail.

**To Reproduce**
on the `x86_64` host install `sqlite3` library and compile for the target `x86`:
```
cmake -B ./build -DAREG_COMPILER_FAMILY=gnu -DAREG_PROCESSOR=x86
cmake --build ./build -j
```
**Expected behavior**
The build should succeed. 
- **SQLite3**: Try to detect the available libraries / packages for the target architecture. If not found, build from sources.
- **ncurses**: Force to disable the build with extended objects if `ncurses` is not available for the target architecture.
- **GTest**:: Force to fetch the sources and build `gtest` library if it is not available for the target architecture.

aregtech:
Fixed the issue with the wrong format of file. This was related with the binaries compiled for different processor targets. The trigger was in `find_package` calls and the root cause was wrong library architecture.
This fixe still needs to be improved.


vukihub (approved):
approving

PR #456, opened 2024-11-29
Fixed the bug, related with copy memory in i386 platform

### LINKED ISSUE #455: StringTestTrim unit test failes
**Describe the bug**
One unit test fails with error message
```
/mnt/c/projects/areg-sdk/tests/units/NEStringTest.cpp:485: Failure
Expected equality of these values:
  NEString::compare(buffer, param.trimRight.data())
    Which is: 1-byte object <FF>
  NEMath::eCompare::Equal
    Which is: 1-byte object <00>

/mnt/c/projects/areg-sdk/tests/units/NEStringTest.cpp:490: Failure
Expected equality of these values:
  NEString::compare(buffer, param.trimAll.data())
    Which is: 1-byte object <FF>
  NEMath::eCompare::Equal
    Which is: 1-byte object <00>

[  FAILED  ] NEStringTest/StringTestTrim.TrimStrings/11, where GetParam() = 36-byte object <0A-00 00-00 14-FD 6E-56 0A-00 00-00 14-FD 6E-56 09-00 00-00 55-FE 6E-56 09-00 00-00 55-FE 6E-56 04-00 00-00> (0 ms)
```
**To Reproduce**
1. compile the sources with following option (fox x86 platform):
   ```
   cmake -B ./product/cache/gnu-linux-x86 -DAREG_COMPILER_FAMILY=gnu -DAREG_PROCESSOR=x86 -DAREG_EXTENDED=ON
   ```
2. run unit tests:
   ```
   ./product/build/gnu-g++/linux-32-i386-release-shared/bin/areg-unit-tests.elf
   ```
**Expected behavior**
Unit tests should run.

**NOTE**: the unit tests pass if compile with option `-DAREG_PROCESSOR=x64`

- OS: Ubuntu Linux
- Hardware: x86
- Compiler: g++
- AREG SDK version: latest

aregtech:
It seems that the `memcpy` method of i386 platform contains a bug. The bug is happens if a memory copied within same buffer, i.e. like moving memory. This happens when "trim" whitespace in the string. Because of this, stopped using copy method and instead implemented copying data in the pointer.

### DIFF
--- a/framework/areg/base/NEString.hpp
+++ b/framework/areg/base/NEString.hpp
@@ -1678,12 +1678,11 @@ void NEString::trimAll( CharType * strBuffer, NEString::CharCount strLen /*= NES
 
             if (strLen != NEString::COUNT_ALL)
             {
-                NEString::copyStringFast<CharType>(buf, next, NEString::COUNT_ALL);
+                while (*next != static_cast<CharType>(NEString::EndOfString))
+                    *buf ++ = *next ++;
             }
-            else
-            {
-                *buf = static_cast<CharType>(NEString::EndOfString);
-            }
+
+            *buf = static_cast<CharType>(NEString::EndOfString);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


vukihub (approved):
looks good

PR #461, opened 2024-12-02
460 fix areg sdk demo workflow

### LINKED ISSUE #460: Fix areg-sdk-demo workflow
**Describe the bug**
The `areg-sdk-demo` has problem to run cross-compiling action workflow.
It finds wrong `sqlite3` library when try to find `SQLite3` package for x86 build. Instead of 32-bit library, it gets 64-bit.
The problem was not able to reproduce locally, but it happens on the GitHub build.

**To Reproduce**
Rung GitHub actions for `areg-sdk-demo`. Run
```cmake
cmake -B ./product/cache/gnu-x86-so -DAREG_COMPILER_FAMILY=gnu -DAREG_PROCESSOR=x86
```
**Expected behavior**
It should find the right library or need to ignore sqlite3 library and build from sources.

**GitHub action:**
- OS: Ubuntu Linux

aregtech:
Implemented macro that dumps binary module to check compatibility with the specific processor architecture.


vukihub (approved):
Looks fine.

PR #469, opened 2024-12-09
463 cannot start mcrouter service

### LINKED ISSUE #463: Cannot Start mcrouter service
**Describe:**
Neither `mcrouter`, nor `lobcollector` can start as a service on Windows platform. The initialization succeeds, the service is listed in the **Service** application, but when try to start, after some period of time, get error message (with timeout).

The application was tested on Windows machine. When fix, it should be tested on Linux as well.

**Hint**: One of possible issues could be that the application starts either without `--service` command line option or it is missed.

aregtech:
* Fixed calling to start service dispatcher (win32 API)
* Updated API to fit the requirement


vukihub (approved):
Looks good.

vukihub (approved):
.

PR #470, opened 2024-12-09
Add new DateTimeTest testcase

biwiki:
Add new DateTimeTest testcase


biwiki:
@aregtech this is the testcase I added while playing with the unittests, not sure if it is worth to merge in

aregtech (approved):
Thank you! Good case.

PR #472, opened 2024-12-11
466 areg build tests

### LINKED ISSUE #466: Question: potential bug while testing AREG_BUILD_TESTS
Hi @aregtech

I am not familiar with cmake, but there seems to be a bug in `conf/cmake/setup.cmake` while testing AREG_BUILD_TESTS

```
if (NOT AREG_BUILD_TESTS)
    set(AREG_GTEST_PACKAGE OFF)
endif()
```

Although, I could't find any information in the main README how to build/run/disable unittest

BR /biwiki

aregtech:
@biwiki ,
Here I've updated as you recommended:
1. In the README of areg-sdk project added description how to exclude unit tests and/or examples;
2. Added separate document with the description to exclude, build and run tests and examples.
3. Other slight modifications.

Could you please review and let me know whether it is clear now?

### DIFF
--- a/docs/wiki/08a-examples-and-tests.md
+++ b/docs/wiki/08a-examples-and-tests.md
@@ -1,0 +1,162 @@
+# Building and Running AREG SDK Examples and Unit Tests
+
+The AREG SDK provides a comprehensive set of examples and unit tests to demonstrate its features and ensure reliability. This guide explains how to include or exclude Examples and Unit Tests during the build process, as well as how to execute them after building.
+
+---
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Prerequisites](#prerequisites)
+- [Configuring the Build Process](#configuring-the-build-process)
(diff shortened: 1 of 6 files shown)


docs/wiki/08a-examples-and-tests.md
```diff
@@ -0,0 +1,161 @@
...
+
+---
+
+## Running AREG SDK Examples
+
+To execute Examples:
+
+1. Navigate to the `bin` directory inside `./product/build/...`.
+2. Locate the desired example binary.
+3. Run the binary from the terminal.
```
biwiki, line 104: I think the standard way to run the test is via cmake build system, like:
```
cd build/
make test
```
But then I am not sure how to pass the user options to the test program

biwiki (approved):

docs/wiki/08a-examples-and-tests.md
aregtech: or (run with filter):
```
ctest --test-dir ./build -R DateTimeTest
``

PR #480, opened 2024-12-21
Update 07a-troubleshooting-wsl-update.md

### LINKED ISSUE #479: Document describing IP address config for Windows and WSL
**Description:**
Due to an issue #336 , there is a need to update "Troubleshooting" document and add a section explaining how to setup the system that components communicate running in Windows and WS platforms at the same.

amesserla:
Add description of IP configuration for Windows and WSL

### DIFF
--- a/docs/wiki/07a-troubleshooting-wsl-update.md
+++ b/docs/wiki/07a-troubleshooting-wsl-update.md
@@ -145,15 +145,15 @@ Additionally, the services on WSL must be provided the correct IP address to con
 
 2. In the WSL configuration file, [``areg.init``](framework/areg/resources/areg.init), the address for any services running as a router or server must be changed to the IP address found above. For example, to connect to an instance of ``mcrouter`` and ``logcollector`` running on Windows, the ``areg.init`` should have the lines
     ```plaintext
-      router::*::address::tcpip = NEW_IP
+      router::*::address::tcpip = <windows-ip-address>
    ```
 
    and 
    ```plaintext 
-      logger::*::address::tcpip   = NEW_IP
+      logger::*::address::tcpip   = <windows-ip-address>
    ```
 
-   changed so that ``NEW_IP`` is replaced with the IP address found above.
+   changed so that ``<windows-ip-address>`` is replaced with the IP address found above.
 
 3. Windows sees the WSL network as being an external network, so the IP address for any services running as a router or server must be set to an IP address which can accept connections from your LAN, such as ``0.0.0.0``. For example, to configure instances of ``mcrouter`` and ``logcollector`` running on Windows to accept connections from services on WSL, the Windows copy of ``areg.init`` for these services should have the lines
    ```plaintext
@@ -166,8 +166,6 @@ Additionally, the services on WSL must be provided the correct IP address to con
 
    with ``0.0.0.0`` or an IP address that is configured to accept external connections replacing the defualt ``localhost``.
 
->[!TIP]
-> For more information about configuring network communication between WSL and Windows, this page may be helpful: [Accessing network applications with WSL](https://learn.microsoft.com/en-us/windows/wsl/networking).
 
 ### Solution 2: Running some servers on WSL and some on Windows
 
@@ -184,3 +182,6 @@ It is possible to mix which services are running on WSL and Windows, but [``areg
 4. Lunch ``logobserver`` and start AREG-based applications on either Windows, WSL, or a mix of both.
 
 The applications will start communicating as intended.
+
+>[!TIP]
+> For more information about configuring network communication between WSL and Windows, this page may be helpful: [Accessing network applications with WSL](https://learn.microsoft.com/en-us/windows/wsl/networking).
--- a/docs/wiki/07a-troubleshooting-wsl-update.md
+++ b/docs/wiki/07a-troubleshooting-wsl-update.md
@@ -143,28 +143,19 @@ Additionally, the services on WSL must be provided the correct IP address to con
       ip route show | grep -i default | awk '{print $3}'
    ```
 
-2. In the WSL configuration file, [``areg.init``](framework/areg/resources/areg.init), the address for any services running as a router or server must be changed to the IP address found above. For example, to connect to an instance of ``mcrouter`` and ``logcollector`` running on Windows, the ``areg.init`` should have the lines
-    ```plaintext
-      router::*::address::tcpip = <windows-ip-address>
-   ```
+2. In the WSL configuration file, [``areg.init``](framework/areg/resources/areg.init), the address for any services running as a router or server must be changed to the IP address found above. For example, to connect to an instance of ``mcrouter`` and ``logcollector`` running on Windows, the ``areg.init`` should have the lines:
 
-   and 
-   ```plaintext 
-      logger::*::address::tcpip   = <windows-ip-address>
-   ```
+   - ``router::*::address::tcpip = <windows-ip-address>``
+   - ``logger::*::address::tcpip   = <windows-ip-address>``
 
-   changed so that ``<windows-ip-address>`` is replaced with the IP address found above.
+changed so that ``<windows-ip-address>`` is replaced with the IP address found above.
 
-3. Windows sees the WSL network as being an external network, so the IP address for any services running as a router or server must be set to an IP address which can accept connections from your LAN, such as ``0.0.0.0``. For example, to configure instances of ``mcrouter`` and ``logcollector`` running on Windows to accept connections from services on WSL, the Windows copy of ``areg.init`` for these services should have the lines
-   ```plaintext
-      router::*::address::tcpip = 0.0.0.0
-   ```
-   and 
-   ```plaintext 
-       logger::*::address::tcpip   = 0.0.0.0
-   ```
+4. Windows sees the WSL network as being an external network, so the IP address for any services running as a router or server must be set to an IP address which can accept connections from your LAN, such as ``0.0.0.0``. For example, to configure instances of ``mcrouter`` and ``logcollector`` running on Windows to accept connections from services on WSL, the Windows copy of ``areg.init`` for these services should have the lines:
 
-   with ``0.0.0.0`` or an IP address that is configured to accept external connections replacing the defualt ``localhost``.
+   - ``router::*::address::tcpip = 0.0.0.0``
+   - ``logger::*::address::tcpip   = 0.0.0.0``
+
+with ``0.0.0.0`` or an IP address that is configured to accept external connections replacing the defualt ``localhost``.
 
 
 ### Solution 2: Running some servers on WSL and some on Windows
@@ -174,8 +165,8 @@ It is possible to mix which services are running on WSL and Windows, but [``areg
 1. Run ``ipconfig`` on Windows and ``ifconfig`` in the WSL terminal to obtain the IP addresses for the PC running Windows and WSL.
 
 2. Modify the ``areg.init`` file on both Windows and WSL to specify the following:
-- ``router::*::address::tcpip = <windows-ip-address>``
-- ``logger::*::address::tcpip = <wsl-ip-address>``
+   - ``router::*::address::tcpip = <windows-ip-address>``
+   - ``logger::*::address::tcpip = <wsl-ip-address>``
   
 3. Start ``mcrouter`` on Windows and ``logcollector`` on WSL.
 
--- a/docs/wiki/07a-troubleshooting-wsl-update.md
+++ b/docs/wiki/07a-troubleshooting-wsl-update.md
@@ -13,8 +13,8 @@ This document outlines solutions to common issues encountered while updating the
    - [Solution 2: Enable Automatic DNS Resolution](#solution-2-enable-automatic-dns-resolution)
    - [Solution 3: Reset WSL Network Configuration](#solution-3-reset-wsl-network-configuration)
 3. [Issue 3: Configuring communication between Windows and WSL](#issue-3-configuring-communication-between-windows-and-wsl)
-   - [Solution 1: Editing IP address settings](#solution-1-editing-ip-address-settings)
-   - [Solution 2: Running some servers on WSL and some on Windows](#solution-2-running-some-servers-on-wsl-and-some-on-windows)
+   - [Solution 1: Editing all IP address settings (Recommended)](#solution-1-editing-all-ip-address-settings-recommended)
+   - [Solution 2: Updating only necessary IP addresses](#solution-2-updating-only-necessary-ip-addresses)
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
 
 > [!TIP]
 > These solutions should address most update and DNS-related issues encountered with WSL. If problems persist, consult the [official WSL documentation](https://learn.microsoft.com/en-us/windows/wsl/) or seek assistance from the community forums.
+
+---
+
+## Issue 3: Configuring communication between Windows and WSL
```
aregtech, line 126: Could you please include a detailed description of **Issue 3** before presenting the solution? It’s important for users to clearly understand the problem first. For example, you might explain it as:  
 
*"There is no communication between applications running on WSL and the system-managed services, such as `mcrouter` and/or `logcollector`, running on Windows."*  

This context will help users better grasp the issue before moving on to the resolution.

docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
...
+### Solution 1: Editing IP address settings
+
+Due to WSL not being able to access the Windows ``localhost`` IP address, routing services hosted on Windows must be configured differently to allow access by clients hosted on WSL.
+Additionally, the services on WSL must be provided the correct IP address to connect to the Windows router.
+
+>[!IMPORTANT]
+>This assumes WSL is using the default networking mode, NAT. Changing this setting to Mirror may allow ``localhost`` to work as intended, but this has not been tested.]
+
```
aregtech, line 140: If there is public available documentation explaining changing **setting to Mirror**, would be fine include link. Not everyone may understand what is the meaning *changing setting to Mirror*.

docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
...
+> [!NOTE]
+> When connecting a client on Windows to a server or router on WSL, the default ``localhost`` settings will allow the connection.
```
aregtech: This is an excellent example of explaining the issue: services running on WSL with `localhost` settings are accessible, whereas services running on Windows with `localhost` settings are not. I suggest moving this explanation to the top of the document, where the issue is initially described, to provide clear context to the users.

docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
...
+    ```plaintext
+      router::*::address::tcpip = NEW_IP
+   ```
+
+   and 
+   ```plaintext 
+      logger::*::address::tcpip   = NEW_IP
+   ```
+
+   changed so that ``NEW_IP`` is replaced with the IP address found above.
```
aregtech: It might be clearer to replace `NEW_IP` with `<wsl-ip-address>` or `<windows-ip-address>`. This would make it easier for users to understand exactly where and which IP address they need to specify.

docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
...
+3. Windows sees the WSL network as being an external network, so the IP address for any services running as a router or server must be set to an IP address which can accept connections from your LAN, such as ``0.0.0.0``. For example, to configure instances of ``mcrouter`` and ``logcollector`` running on Windows to accept connections from services on WSL, the Windows configuration files for these services should have the lines
+   ```plaintext
+      router::*::address::tcpip = 0.0.0.0
+   ```
+   and 
+   ```plaintext 
+       logger::*::address::tcpip   = 0.0.0.0
+   ```
+
+   with ``0.0.0.0`` or an IP address that is configured to accept external connections replacing the defualt ``localhost``.
```
aregtech: In my case, I conducted the tests as follows:

1. **Windows**: Run `mcrouter` on Windows.  
2. **WSL**: Run `logcollector` on WSL.  
3. Use `ipconfig` on Windows and `ifconfig` in the WSL terminal to obtain the IP addresses of the PC running Windows and WSL.  
4. Modify the `areg.init` file on both Windows and WSL to specify the following:
   - `router::*::address::tcpip = <windows-ip-address>`  
   - `logger::*::address::tcpip = <wsl-ip-address>`  
5. Start `mcrouter` on Windows and `logcollector` on WSL.  
6. Launch `logobserver` and start AREG-based applications on either Windows, WSL, or a mix of both.

**Result**: The applications started successfully, and communication worked seamlessly.  

Could you please also specify this as **Solution 2**?

docs/wiki/07a-troubleshooting-wsl-update.md
```diff
@@ -117,3 +119,49 @@ If the issue persists, reset the WSL network configuration:
...
+> [!NOTE]
+> When connecting a client on Windows to a server or router on WSL, the default ``localhost`` settings will allow the connection.
+
+1. To configure the WSL clients, the IP address of the connection to Windows must be specified. The fastest way to find this IP address is to run the following in the WSL terminal:
+
+   ```bash
+      ip route show | grep -i default | awk '{print $3}'
+   ```
+
+2. In the WSL configuration file, the address for any services running as a router or server must be changed to the IP address found above. For example, to connect to an instance of ``mcrouter`` and ``logcollector`` running on Windows, the configuration file should have the lines
```
aregtech: > *the configuration file*

Could you please specify the exact file name, `areg.init` (probably with the link to the file in repo), so that users clearly know which file needs to be modified? This will help avoid any confusion and ensure accurate updates.

aregtech (changes requested):
@amesserla,  
Thank you very much for updating the document.  
Overall, it looks good. However, a few adjustments could make the document clearer and more helpful.  
Kindly review my notes for further improvement.

amesserla:
I have made the requested changes. Let me know if there are any more improvements you think I could make!

aregtech:
Hi @amesserla,  

Thank you for your contributions! The updates look much better now!  

I’ve reviewed the file and made some updates to further enhance clarity and accuracy. Since I cannot directly push changes to the pull request you’ve created, I’ve created a new branch with the modified file for your review: Modified [07a-troubleshooting-wsl-update](https://github.com/aregtech/areg-sdk/blob/pull/480/docs/wiki/07a-troubleshooting-wsl-update.md) file.  

Key Updates:  
1. Corrected spelling errors.  
2. Updated links to the `areg.init` file.  
3. Refined some wording for better clarity. Specifically, I noticed you mentioned "services and servers" in a few places. In the context of the `areg-sdk`, **services** often refer to **physical** servers. To avoid confusion, I clarified that the AREG Framework supports both **logical** and **physical** servers, enabling software components to act as **logical servers**, **logical clients**, or both. However, this distinction isn’t directly relevant to the physical IP connection discussed here, so I adjusted the phrasing accordingly.  

Could you please review the changes, especially in *Solution 1*, to confirm if everything is accurate? If you agree with the updates, feel free to copy and push them to the pull request.  

Many thanks in advance for your time and efforts!

aregtech (changes requested):
Please compare with the changes in https://github.com/aregtech/areg-sdk/blob/pull/480/docs/wiki/07a-troubleshooting-wsl-update.md

amesserla:
Hi @aregtech, 

I'm away from the computer I test on for a few days, so I can't answer this myself right now. In the re-writing of Solution 2: it appears you don't need to edit the ``router::*::address::tcpip = localhost`` line in ``areg.init`` on Windows. Wouldn't this cause ``mcrouter`` to listen on ``127.0.0.1`` and fail to connect with ``logcollector`` on WSL?

Thank you for clarifying the differences between the usage of services and servers. I'll look carefully at your usage so I can note the differences when looking through other parts of the documentation.

aregtech:
Hi @amesserla,  

No problem at all, there's no rush! We can continue once you're back.  

> Wouldn't this cause mcrouter to listen on 127.0.0.1 and fail to connect with logcollector on WSL?  

That's a good question, and to be honest, I have no clear answer. It seems testing of various configurations is needed. It looks like you have a deeper experience and understanding of WSL and Windows networking than I do. If you could find the time to test multiple variations and document your observations, it would be incredibly helpful. Your insights would definitely contribute to improving this aspect of the project.

amesserla:
Hi @aregtech,

I am back to my pc and did some testing, and here is what I found. 

The setup had `mcrouter` running on Windows, and `logcollector` on WSL, with `logobserver` running one instance on Windows and one instance on WSL. I tested all combinations of the following 4 iterations of `areg.init`.

Config 1 contained:
```
router::*::address::tcpip  = localhost
logger::*::address::tcpip = <wsl-ip>
```

Config 2 contained:
```
router::*::address::tcpip  = <windows-ip>
logger::*::address::tcpip = <wsl-ip>
```

Config 3 contained:
```
router::*::address::tcpip  = localhost
logger::*::address::tcpip = localhost
```

 Config 4 contained:
```
router::*::address::tcpip  = <windows-ip>
logger::*::address::tcpip = localhost
```

Results:

The results are displayed in the following table where the number on top specifies the modification to `areg.init` used on Windows and the left column specifies the modification to `areg.init` used on WSL. Specifically, the 4th row and 1st column represents how I understood the suggested configuration from [https://github.com/aregtech/areg-sdk/blob/pull/480/docs/wiki/07a-troubleshooting-wsl-update.md](https://github.com/aregtech/areg-sdk/blob/pull/480/docs/wiki/07a-troubleshooting-wsl-update.md)

|  | 1 | 2 | 3 | 4 |
|---|---|---|---|---|
| 1 | ✔️ | ✔️ | ❌ | ❌ |
| 2 | ✔️ | ✔️ | ❌ | ❌ |
| 3 | ❌ | ❌ | ✔️ | ✔️ |
| 4 | ❌ | ❌ | ✔️ | ✔️ | 

✔️ - indicates that both instances of `logobserver` showed a connection to `mcrouter` and `logcollector` showed a connection to `mcrouter` and both instances of `logobserver`
❌ -  indicates that neither instance of `logobserver` showed a connection to `mcrouter` and `logcollector` only showed a connection to the instance of `logobserver` running on WSL

In particular, for a successful connection, the line `logger::*::address::tcpip` must match between the copies of `areg.init` on WSL and on Windows, but the line `router::*::address::tcpip` does not need to match. 

I looked back at the original issue of running both `mcrouter` and `logcollector` on Windows, and `logobserver` on WSL and found that the connection works if the line `logger::*::address::tcpip = localhost` is changed to `<windows-ip>` in both `areg.init` files, but changing the line `router::*::address::tcpip  = localhost` is not necessary in either. This same configuration works with `mcrouter` running on WSL and `logobserver` running on either.

-----------

My guess as to what is happening is that the `logcollector` application is functioning as the server that `mcrouter` and `logobserver` are connecting to. If `logcollector` is running on WSL, it can either be accessed by `localhost` (since Windows has access to the WSL `localhost`) or by `<wsl-ip>`, but which ip address is used must be specified to the Windows applications via `areg.init`. If `logcollector` is running on Windows, the ip address for `logcollector` must be changed from `localhost` (because WSL does not have access to the Windows `localhost`). 

I should mention that I have only done testing with `mcrouter`, `logcollector`, and `logobserver`. It may be that sending or receiving information between another application and `mcrouter` (if this is something that can happen) would require other configuration changes. Do you have a recommended way of testing this?

aregtech:
Hi @amesserla .
Happy New Year! :)

Thank you very much for your detailed report—this is truly outstanding work.

About applications communicating via `mcrouter`, I believe it works similarly to communication with `logcollector`. If I understand correctly, the best solution would be to set both `<windows-ip>` and `<wsl-ip>` in the configuration. We could present this as the recommended approach while noting that using `localhost` in the configuration for applications on WSL can serve as an alternative solution.

Could you please incorporate a summary of this into the document? Also, since your report is very detailed and helpful, it would be great to include a permanent link to [your message](https://github.com/aregtech/areg-sdk/pull/480#issuecomment-2568508064). If you're unsure whether GitHub keeps pull requests long-term, you could copy your message to [this ticket](https://github.com/aregtech/areg-sdk/issues/479) and refer to it in the document. Alternatively, you could include the message directly in the document.  

Please feel free to choose the approach you find most appropriate. The goal is to give users clear and detailed information.  

Thanks again for your excellent work!

aregtech:
> ✔️ - indicates that both instances of `logobserver` showed a connection to `mcrouter` and `logcollector` showed a connection to `mcrouter` and both instances of `logobserver`
> ❌ - indicates that neither instance of `logobserver` showed a connection to `mcrouter` and `logcollector` only showed a connection to the instance of `logobserver` running on WSL

Hi @amesserla,  

I wanted to quickly clarify to ensure there’s no misunderstanding:  

1. **`logcollector`** serves as the central server for all applications that are registered as "log sources" to send log messages.  
2. **`logobserver`** is a client of `logcollector` and does not have a direct physical connection to `mcrouter`.  
3. **`mcrouter`** acts as a physical server for all applications but also functions as a client of `logcollector` to send log messages.

amesserla:
Hi @aregtech 

> I wanted to quickly clarify to ensure there’s no misunderstanding:

This helped clarify the different applications for me a lot. 

> About applications communicating via `mcrouter`, I believe it works similarly to communication with `logcollector`

I did some more testing to confirm this, using the [`20_pubdatarate`](https://github.com/aregtech/areg-sdk/tree/master/examples/20_pubdatarate) example to confirm that if applications running on WSL want to communicate with `mcrouter` running on Windows, then both `areg.init` files (on WSL and Windows) must be updated to specify `router::*::address::tcpip  = <windows-ip>`. This fits into my understanding that if an application on WSL wants to make a connection with a server on Windows, acceptable ip addresses must be specified.

Specifically:

I launched `mcrouter` and `20_pubclient` on Windows and `20_pubserver` on WSL and started `20_pubserver`. When the default `router::*::address::tcpip  = localhost` was left unchanged in either copy of `areg.init`, no connection occurred. However, after changing  `router::*::address::tcpip  = <windows-ip>` in both `areg.init` files, the connection occurred and data was exchanged as intended.

With `mcrouter` running on WSL and `20_pubclient` and `20_pubserver` on Windows , using `localhost` in both `areg.init` files, or using `<wsl-ip>` in both files, worked as intended.

> the best solution would be to set both `<windows-ip>` and `<wsl-ip>` in the configuration.

I agree that this is the best solution. If it sounds good to you, I will rewrite the documentation to offer this as solution 1, and use my explanations above to offer a solution 2 which includes the fact you can use `localhost` for applications running as a server on WSL, even if clients running on Windows want to connect to that server, but with the caveat that the specified ip addresses must match between the two copies of `areg.init`.

aregtech:
Hi @amesserla 

> If it sounds good to you, I will rewrite the documentation

Yes please, this sounds absolutely fine.

aregtech (approved):
@amesserla , thank you very much. It is informative and good described.

PR #483, opened 2025-01-05
fixup libs var in unittest makefile

biwiki:
_libs variable name is incorrectly handled in this makefile, I think

### DIFF
--- a/tests/units/CMakeLists.txt
+++ b/tests/units/CMakeLists.txt
@@ -17,6 +17,7 @@ macro(macro_add_unit_test test_project)
 
     set(_tests)
     set(_list "${ARGN}")
+    set(_libs)
 
     foreach(item IN LISTS _list)
         list(APPEND _tests "${AREG_UNIT_TEST_BASE}/${item}")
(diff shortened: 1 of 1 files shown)


tests/units/CMakeLists.txt
biwiki, line 1: I also think that tests should not link to areg library, because that will cause us problems when creating mocks

aregtech: Hi @biwiki

Thank you very much for your fix!

> I also think that tests should not link to areg library.

How then we can make test `areg` library? Do you suggest to link compiled `obj` files?

biwiki: It will not a straight forward task, but for me tests should divided into multiple small ones, each test will contain test class(es), mock class(es) and a subset of classes in production code which is the areg library

aregtech: @biwiki, 

To be honest, I’m having difficulty understanding how you envision the proposed code structure. Could you provide a small working example to demonstrate your approach? Using the existing unit tests as a basis for comparison would be perfectly fine. It is to compare the existing solution with your suggestion.

I’m absolutely open to improvements, but since I don’t fully grasp how you intend to organize or structure the code, I’m unable to provide a clear answer at this point.

aregtech (approved):
@biwiki thank you for the fix.
The discussion moved [here](https://github.com/aregtech/areg-sdk/discussions/485)

PR #500, opened 2025-03-27
499 save logs in the custom database file

### LINKED ISSUE #499: Save logs in the database file
**Description:**

There should be possibility to save logs in the sqlite database file with the custom path bypassing `areg.init` file.
The location in `areg.init` file should be used only if the path is not given.

aregtech:


vukihub (approved):
looks fine

PR #505, opened 2025-04-05
Implement saving configuration

### LINKED ISSUE #503: Change logobserver IP-address and save configuration
**Description:*

There should be possibility during runtime to change the following attributes of `logobserver`:
- IP-address of the logobserver
- TCP/IP port number of the logobserver
- The logging file name
- Some other attributes (???)

After change, there should be possibility to save changed data in the `areg.init` configuration file.

aregtech:
Implement modifying and saving configuration.
added new global API and extended log observer base class.


vukihub (approved):
looks fine.

PR #508, opened 2025-04-07
Add reference counter to the log observer

### LINKED ISSUE #507: Reference counter for log observer
**Description:**

Implement reference counter when initializing and releasing the log observer object (`areglogger`).
The log observer should be initialized when called first time.
Each time it is called to initialize, it should increase the reference counter.
Each time it is called to release, it reduces the reference counter.
The lob observer is released then the reference counter is zero.

aregtech:
Added reference counter to the log observer, which is increased each time the log observer is initialized and decreased when released. The log observer is stopped when log observer reference counter reaches 0.

### DIFF
--- a/framework/areglogger/client/private/LogObserverApi.cpp
+++ b/framework/areglogger/client/private/LogObserverApi.cpp
@@ -39,12 +39,17 @@ namespace
 {
     struct sLogObserverStruct
     {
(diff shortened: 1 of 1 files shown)


vukihub (approved):
looks fine.

PR #513, opened 2025-04-22
fixed 512 cannot load unloaded model

### LINKED ISSUE #512: Cannot load unloaded model
**Description:**

The system cannot load unloaded model second time.

**Reproduction:**
1. Create a model --> unique model is created;
2. Load the model --> model is loaded and the component is created;
3. Unload the model --> component is deleted and the model is unloaded.
4. Load the model again

**Result:**
On step 4, the model is not loaded and the component is not created.

*Hint:* Most probably, the unloaded model is not marked as unloaded, so that when the system tries to load the model and register again, it fails, because the system assumes that the model with the name is already loaded and component are instantiated.

aregtech:
- Fixed a bug to load unloaded model;
- Fixed bugs in log collector client object.


vukihub (approved):
looks good.

PR #515, opened 2025-04-25
514 log database location and name

### LINKED ISSUE #514: Log database location and name
**Description:**

There is a mismatch when creating log database with C-style API or log collector client API. In case of C-style API, the engine is using `db location` to create the log file. In case of log collector client -- the log file is created by combination `db location` and `db name`. The `db name` property set in the `areg.init` file is used to indicate the database engine. So that, there should be a fix:

1. The `db location` is indicating the directory of logs. If it is empty, the default location is in subdirectory `./logs/`
2. The `db name` is indicating the name of the log file, the default is `log_%time%.sqlog`
3. There should be additional property `db engine` indicating the database driver such as `SQLite3`.

There must be no difference using C-style API or log collector client object. Both should act in the same way.

aregtech:
Fixed bug related with the location and database file name.

### DIFF
--- a/framework/areg/base/File.hpp
+++ b/framework/areg/base/File.hpp
@@ -583,6 +583,18 @@ public:
      **/
     static int splitPath(const char * filePath, StringList & in_out_List);
 
+    /**
+     * \brief   From given directory name and file name creates the full path of the file.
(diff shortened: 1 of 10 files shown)


vukihub (approved):
looks fine

PR #527, opened 2025-06-03
Was not able to reconnect to log observer.

### LINKED ISSUE #526: Logobserver failes to reconnect
**Describe the bug**

The `areglogger` as a base of logobserver is used in the `lusan` application.
If the user connects first time, everything looks fine: the scopes are registered, the log messages are received, the log priority is changing, i.e. `lusan` communication with `logcollector` service.

But if the user clicks on disconnect to disconnect the service and tries to connect again, it cannot connect to the `logcollector` anymore.

**To Reproduce**
1. Start `logcollector` and `mcrouter`, for example;
2. Start `lusan` (`areg-sdk-tools`).
3. In the navigation window of `lusan`, go to the `Live Logs` and click on `Connect` toolbutton
    - The scopes are registered and displayed;
    - The logs are received.
4. Click to `Disconnect` button (same as `Connect`) to disconnect from `logcollector`
    - The scope navigation window shows no scopes;
    - Logs are not received anymore.
5. Click on `Connect` button again:
    - The scope navigation window does not show scopes anymore;
    - The logs are not received anymore;
    - The `Connect` button does not indicate connection.

**Expected behavior**
It should be possible always to connect and disconnect `logcollector`.

**Additional context**
It seems that when click on `Disconnect` to disconnect from `logcollector`, the _Send Message_ thread exits, so that it does not reconnect anymore.

aregtech:
The problem was in the conflict of the thread names. Both, `areglogger` and `lusan` application where using `LogObserverThread` name to create threads. Fixed by adding prefix to the thread name.

### DIFF
--- a/framework/areglogger/client/private/LoggerClient.hpp
+++ b/framework/areglogger/client/private/LoggerClient.hpp
@@ -62,7 +62,7 @@ class LoggerClient  : public    ServiceClientConnectionBase
 //////////////////////////////////////////////////////////////////////////
 private:
     //!< The name of the observer dispatcher thread
-    static constexpr std::string_view                   ThreadName      { "LogObserverThread" };
+    static constexpr std::string_view                   ThreadName      { "AREG_LogObserverThread" };
 
     //!< The prefix to add to the send and receive threads.
     static constexpr std::string_view                   ThreadPrefix    { "Observer" };


vukihub (approved):
ok

PR #531, opened 2025-06-16
Fixed returning log database file name instead of returning full path.

### LINKED ISSUE #530: Wrong log database file name
**Description:**
The method `LoggerClient::getConfigDatabaseName()` instead log database file name set in the configuration file returns the full path to the database file, where the database file name contains a mask.

**To Reproduce**
- Initialize cogger client object `LoggerClient`.
- Call `LoggerClient::getConfigDatabaseName()` function.

**Expected behavior**
The method `LoggerClient::getConfigDatabaseName()` should return only file name set in the configuration file. Meaning if the file contains a mask like `%time%.sqlog`, it should return this value instead of full path.

aregtech:
Fixed returning log database file name instead of returning full path.

### DIFF
--- a/framework/areglogger/client/private/LoggerClient.cpp
+++ b/framework/areglogger/client/private/LoggerClient.cpp
@@ -329,7 +329,7 @@ String LoggerClient::getConfigDatabaseName(void) const
     if (isSqliteEngine())
     {
         LogConfiguration config;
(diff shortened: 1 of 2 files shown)


vukihub (approved):
Looks fine. It should be tested more.

PR #534, opened 2025-06-21
533 check localhost address

### LINKED ISSUE #533: Check localhost address
**Description:**

The `localhost` and `127.0.0.1` are identical. Add a method in `NESocket` namespace to check localhost name or IP

aregtech:
- added checkup of local host;
- added validator of IPv4 address as a string
- modified SocketAddress and added possibility to keep IP address and host name.

### DIFF
--- a/framework/areg/base/private/Socket.cpp
+++ b/framework/areg/base/private/Socket.cpp
@@ -116,7 +116,7 @@ int Socket::receiveData( unsigned char * buffer, int length ) const
 
 bool Socket::setAddress(const char * hostName, unsigned short portNr, bool isServer)
 {
-    if ( isValid() && mAddress.isEqualAddress(hostName, portNr) )
+    if ( isValid() && (mAddress.isEqualAddress(hostName, portNr) == false))
     {
         ASSERT(mSocket.get() != nullptr);
         decreaseLock( );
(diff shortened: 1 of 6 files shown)


framework/areg/base/private/Socket.cpp
```diff
@@ -116,7 +116,7 @@ int Socket::receiveData( unsigned char * buffer, int length ) const
 
 bool Socket::setAddress(const char * hostName, unsigned short portNr, bool isServer)
 {
-    if ( isValid() && (mAddress.getHostAddress() != hostName || mAddress.getHostPort() != portNr) )
+    if ( isValid() && mAddress.isEqualAddress(hostName, portNr) )
```
vukihub: fix the bug
```cpp
    if ( isValid() && (mAddress.isEqualAddress(hostName, portNr) == false))
    {
        ASSERT(mSocket.get() != nullptr);
        decreaseLock( );
    }
```

framework/areglogger/client/private/LogObserverBase.cpp
```diff
@@ -109,7 +119,7 @@ bool LogObserverBase::getConfigLoggerEnabled(void) const
     return logObserverConfigLoggerEnabled();
 }
 
-std::string LogObserverBase::getConfigLoggerAddress(void) const
+const std::string LogObserverBase::getConfigLoggerAddress(void) const
```
vukihub: probably we don't need here `const` modifier.

vukihub (changes requested):
Please read comments.

vukihub (approved):
looks fine

PR #537, opened 2025-06-28
Updates of SQLite wrappers.

### LINKED ISSUE #535: Extend sql database object to fetch data
**Description:**

Currently the SQLIte database object of extended AREG framework library can execute SQL. It is used to insert the logging data. But there are no methods to fetch data from database. The library should be extended to be able to extract various logging data. It can be used to set and run filters.

**Implementation:**

The implementation of extracting data could be done in `LogSqliteDatabase` object.

aregtech:
Added new SQLite wrapper classes to fetch data or run queries.


vukihub (approved):
looks fine

PR #540, opened 2025-07-03
539 enable open database

### LINKED ISSUE #539: Enable open database
**Description:**
Currently when open database (`connect()`) it also creates tables and indexes, i.e. initializing database tables.
There should be a possibility to open existing database without initialization.
Ideally, it should check whether the database file exists. If exists, open without initialization. If does not exist, create database, create tables and indexes.

aregtech:
Enabled opening database for read-only mode.
Made slight optimizations


vukihub (approved):
looks fine.

PR #547, opened 2025-07-14
546 log entry count

### LINKED ISSUE #546: Log entry count
**Description:**

Add method to count number of entries and number of scopes.
Modify methods not only to append entries, but set entries if the size of array is bigger than the index to add to the array. Use new method for that to indicate the current position.

aregtech:
Implemented methods to read log, scopes and log instances count.
Implemented method to set log, scopes and instance information in the array without resizing it.


vukihub (approved):
looks fine.

PR #549, opened 2025-07-23
548 session for logs

### LINKED ISSUE #548: Session for logs
**Description:**
The logs should have sessions. Each time the scope is initialized, the session number of the should be increased. So that, the logs can be grouped not only by scopes, but also by sessions.

aregtech:
Implemented session concept for the logs of the same scope.
Integrated and tested with `lusan` (areg-sdk-tools) project.


vukihub (approved):
Looks fine.

PR #551, opened 2025-07-25
550 assertion on sql statement when close application

### LINKED ISSUE #550: Assertion on SQL statement when close application
**Description**

When closed `lusan` application, have got the assertion failure to log message.
This happens sporadic.
The root cause might be in multithreading environment when close database and invalidate the statement object, and meantime receiving a log message.

The assertion is raise in the `LogSqliteDatabase` then calling `logMessage()`.
```cpp
bool LogSqliteDatabase::logMessage(const NELogging::sLogMessage& message, const DateTime& timestamp)
{
    if (mStmtLogs.isValid() == false)
    {
        ASSERT(false && "LogSqliteDatabase::logMessage() - The statement to log messages is not prepared.");
        return false;
    }
...
}
```

And the caller is
```cpp
void ObserverMessageProcessor::notifyLogMessage(const RemoteMessage& msgReceived)
{
...
        mLoggerClient.mLogDatabase.logMessage(*msgRemote, now);
        mLoggerClient.mLogDatabase.commit(true);
...
}
```

**Solution:**
It worth to check the statement object validity and escape committing database when logging message.

aregtech:
Running logging SQL is now thread safe. Considered cases when the database can be closed and operable.

### DIFF
--- a/framework/aregextend/db/LogSqliteDatabase.hpp
+++ b/framework/aregextend/db/LogSqliteDatabase.hpp
@@ -348,7 +348,7 @@ public:
      * \param   maxEntries  [in]    The maximum number of entries to extract. If `-1`, it extracts all entries.
      * \return  Returns number of entries added to the vector.
      **/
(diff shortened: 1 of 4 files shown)


vukihub (approved):
looks fine

PR #553, opened 2025-07-26
552 fetching instances goes in loop

### LINKED ISSUE #552: fetching instances goes in loop
**Description**

Fetching log instances goes in the loop.
Wrong implementation of `next()`.

aregtech:
fixed using fetching data to escape loop.


vukihub (approved):
Is OK

PR #554, opened 2025-07-26
Fixed assertion, increased auto-typing timeout

### LINKED ISSUE #541: Assertion raised when close window in example 17
**Description:**
1. Start win chatter (example 17)
2. Trigger automatic chat (generate letters);
3. While application runs, close any window;

**Result:**
Assertion is raised on the windows that are connected with closed application:

Image

aregtech:
Fixed the assertion and increased the auto-typing timeout, because in windows it makes no set 1 ms timeout. In Windows it makes sense to set timeout 10 ms and more.


vukihub (approved):
it's OK.

PR #559, opened 2025-07-30
558 set received time in logs

### LINKED ISSUE #558: Set received time in logs
**Description:**
In the `NELogging::sLogMessage` structure add a new entry to set the received or updated timestamp, which should be possible to visualize on log observer side (`lusan`) application. This can be an additional helpful information to handle in case if a process or network is overloaded and cannot process log messages on time. The user at least can compare and measure times when the log message is generated and when it is received in the log observer side.

aregtech:
Added received log datetime in the logs structure. Updated instance connection and scope registration mechanism to log messages to notify the event.


vukihub (approved):
it's fine.

PR #561, opened 2025-07-30
560 cleaning log list case exception

### LINKED ISSUE #560: Cleaning log list case exception
**Description:**

- Start `lusan` application (`areg-sdk-tools`);
- Open the logs in offline mode;
- Close the log viewer window.

**Result:** the application is crashing referring to the `SharedBuffer` destructor indicating corrupted heap.
The bug is most probably in `areg-sdk`, but it should be more investigated.

aregtech:
Fixed initializing log message to store data in database correctly.
Otherwise, BLOB was saved in database and this was corrupting allocated buffer.


vukihub (approved):
looks fine

PR #584, opened 2025-09-07
567 rename mcrouter to mtrouter

### LINKED ISSUE #567: Rename mcrouter to mtrouter
**Naming: “multicast” vs “multitarget”**

* **Physical behavior: It routes one message to multiple targets. That is multitarget.**
* **Industry terminology:**
  * “Multicast” is widely understood in networking contexts, but implies network layer multicast, which might mislead readers into thinking it uses IP multicast or specific network protocols.
  * “Multitarget” is less common as a standard term, but it is more precise for what areg does: route to multiple registered service consumers or providers.
* **Developer clarity test:**
  * A C++ dev seeing “multicast router” might assume: “Is this using UDP multicast? Will it conflict with firewall settings?” → confusion.
  * A C++ dev seeing “multitarget router” will think: “It can deliver to multiple endpoints” → accurate mental model.“Multitarget router” is the better term for README and docs. It minimizes confusion with IP multicast and describes exactly what the router does in areg’s architecture. We can still parenthetically note “(acts like a multicast router for logical service nodes)” the first time it appears, so readers who know that paradigm connect the dots.

**Conclusion**
* Rename the module _mcrouter_ --> _mtrouter_
* Fix everywhere _Multicast router_ --> _Multitarget router_.

aregtech:
- renamed `multicast` to `multitarget`
- renamed `mcrouter` to `mtrouter`
- fixed the paths and file names.


vukihub (approved):
looks fine

PR #589, opened 2025-09-11
Made optimization of registering services

### LINKED ISSUE #588: use functions instead of pointers to function
**Description:**

In most of the example codes and in real live, the servicing components are created in the same way. The models are using the pointers to the component objects to instantiate the servicing components. Use more modern way to define methods that create components, instead of pointers to functions use `std::function`, instead of static methods to instantiate servicing components use default lambda function lie this `[&](const NERegistry::ComponentEntry& entry, ComponentThread& owner) -> Component*{ return DEBUG_NEW ServiceProvider(entry, owner);}`

The same is Delete method.  
Make sure the code compiles, the model is solid, the objects are created and deleted.

aregtech:
* Removed pointer to method to create and delete components;
* Since all calls to create and delete components are similar, used `std::function` and lambda to create object;
* Integrated new changes in the codes and updated example to remove unused methods.


vukihub (approved):
looks very good

PR #590, opened 2025-09-13
Fixed example 20. Optimized using any value in the component registry.

### LINKED ISSUE #585: Fix example 20
**Description:**

The example 20 is compiling, but not working. Should be fixed.
It runs only on windows.
Compiles with clang-cl compiler on command line.

aregtech:
* optimized using any data in the component registry;
* fixed example 20


vukihub (approved):
looks fine

PR #594, opened 2025-09-21
Example of using areg event synchronization object and STL threads

### LINKED ISSUE #593: Example of synch event in STL
**Description:**
Create a new example of using event synchronization object with STL object to prove compatibility of areg framework with STL.

aregtech:
Added new example to demonstrate the event synchronization object interacting in STL based thread.


vukihub (approved):
looks fine

PR #600, opened 2025-09-28
Small fix for posix platforms.

dyfet:
Actually sys/unistd.h is not present on some posix targets, and when it's not, you also get errors with sys/signal.h. Tgis uses header checl to see if sys/unistd.h exists and falls back to <unistd.h>. Depending on what is done here, it might be safer to use <csignal> ubstead. Or maybe <unistd.h> / <signal.h> should be the default.  Hard to say since I may not test on the same platforms you have, so the header check seemed safest.

Certaiinly with this change it builds straight out of the box on my alpine main branch, and all 359 unit tests do pass. Since its a header conditional (C++17) it should not build differently elsewhere where it used to work:

100% tests passed, 0 tests failed out of 359

Total Test time (real) =   3.93 sec

### DIFF
--- a/framework/areg/base/private/posix/ThreadPosix.cpp
+++ b/framework/areg/base/private/posix/ThreadPosix.cpp
@@ -27,13 +27,19 @@
 #include <limits.h>
 #include <pthread.h>
 #include <sched.h>
-#include <time.h> 
+#include <time.h>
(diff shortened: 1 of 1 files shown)


aregtech (approved):
Looks fine. Thank you very much!

PR #614, opened 2025-10-23
Increased the block size of shared buffer.

### LINKED ISSUE #611: Configurable memory alignment block size
**Description:**
In the current version of the Areg Framework, buffer allocations are aligned to `sizeof(uAlign)`, which equals **8 bytes (64 bits)**. While this works well in general, **frequent dynamic memory allocations**—especially in small systems with limited RAM—can lead to **heap fragmentation** and degraded performance.

To mitigate this, introduce the ability to **configure the minimal allocation block size dynamically**. This block size will also determine the **memory alignment unit** used for buffer allocations.

**Requirements:**

1. The **minimal alignment size** must always be **at least 8 bytes (64 bits)**.
2. The **alignment block size** should be **configurable via the framework’s configuration file**, allowing adjustment without recompilation.
3. The **default alignment size** can be, for example, **64 bytes (8×8)** or **128 bytes (16×8)**.
4. The **maximum configurable block size** should be limited to **4 KB**.
5. When the application starts and loads the configuration, the framework should automatically use the configured alignment size for all dynamic memory allocations handled by the internal buffer manager.

**Goal:**
Improve runtime memory efficiency and reduce fragmentation on resource-constrained systems by allowing projects to **tune memory alignment behavior at startup** instead of compile time.

aregtech:
* Added configuration entry in `areg.init` to set the buffer block size;
* Made appropriate changes to read block size from config;
* Prepared the SharedBuffer to set configurable data, but did not integrate, because this might influence on performance. Need to make measurements.


framework/areg/persist/ConfigManager.hpp
```diff
@@ -758,6 +758,12 @@ class AREG_API ConfigManager
      **/
     void setLogDatabaseProperty(const String & whichPosition, const String & newValue, bool isTemporary = false);
 
+    uint16_t getDefaultBufferBlockSize(const String& whichModule = NEString::EmptyStringA);
```
vukihub, line 766: Document the method

vukihub (approved):
looks fine

PR #615, opened 2025-10-25
Add possibility to set stack size of thread.

### LINKED ISSUE #608: Configurable stack size of the thread
**Description:**

In embedded systems, threads are often memory-expensive because each one allocates a fixed amount of stack space.  The default stack size varies by platform — for example, Windows uses 1 MB per thread, while many embedded systems default to around 64 KB.  

- Simple threads often don’t need that much memory, leading to wasted RAM.  
- Some threads may require *more* than the default stack size, causing instability or unexpected behavior.

To improve memory efficiency and flexibility, AREG SDK should support **customizable thread stack sizes**.

**Proposed Solution:**
- Allow specifying a thread’s stack size at creation time, e.g., via the thread constructor or a configuration parameter.  
- Allow defining a **global default minimum stack size** for all threads in the process.  
  - This global default applies only to threads created *after* it is set.  
  - Existing threads remain unaffected.

> [!NOTE]
> Stack size configuration must occur **before** the thread starts.  
> Changing the default stack size affects only *future* threads, not those already created.

aregtech:
Set the stack size of the thread or ignore setting stack size.


vukihub (approved):
looks ok

PR #632, opened 2025-12-11
disable filtering mask 

### LINKED ISSUE #631: Disable log priority
**Description:**

In the SQLite log object add functionality to disable logs for the certain target ID. Disabling logs, means that it is temporary set to "PrioNotset". This will disable logs from a particular instance ID.

**Solution:**
Add new method to call an SQL to set filter flag `0x01` (`PrioNotset`).

aregtech:
## Summary

Added function to disable filtering mask in the log file of a specific instance and all instances

## Changes
Added 2 SQL and a method.

## Testing

Lusan project

## Checklist

Signed-off-by: Artak Avetyan


vukihub (approved):
It is fine

PR #645, opened 2026-01-03
Fixing issue with resource map.

### LINKED ISSUE #644: Exception at Resource Map
**Description:**

The resource map of worker threads sometime throws an exception. The reason is that it tries to find an entry in the empty resource map object.

This is clearly visible in the `areg-edgeai` project when start AI Agent app, activate model and then close the app. As a rule, it throws an exception.

aregtech:
## Summary
* Fixed issue with the resource map;
* Extended Stub Event consumer API to save currently processing event object.

Signed-off-by: Artak Avetyan


vukihub (approved):
looks OK

PR #675, opened 2026-02-17
Add comprehensive unit tests for String class (#122)

### LINKED ISSUE #122: Unit Tests for strings

apoorvdarshan:
## Summary

- Add 138 unit tests for the `String` class in new `tests/units/StringTest.cpp`
- Cover constructors, assignment, comparison, concatenation, search, substring, manipulation, trim, case conversion, numeric conversions, format, validation, and edge cases
- Add `StringTest.cpp` to `tests/units/CMakeLists.txt` build list

Closes #122

## Notes on library bugs discovered during testing

- `replace(char, char)` with `replaceAll=true` has an infinite loop when non-matching characters are encountered (missing `else { ++dst; }` in `TEString.hpp`)
- `trimLeft(TEString&)` copy variant crashes due to using `strResult.length()` (0 after clear) instead of `mData.length()` in `TEString.hpp`
- `findLast(phrase)` has an off-by-one in its `startPos` calculation (`strLen - 1 - count` should be `strLen - count`)

Tests were written to work around these existing issues.

## Test plan

- [x] Build: `cmake --build build --target areg-unit-tests`
- [x] Run: `./build/bin/areg-unit-tests --gtest_filter="StringTest*"` — 138 tests pass
- [x] Full suite: all 497 tests pass (no regressions)

## Checklist

- [x] I have read CONTRIBUTING.md

Signed-off-by: Apoorv Darshan <apoorvdarshan@gmail.com>


aregtech:
@apoorvdarshan thank you very much for this PR. I'll check and merge. But it will take a little time, because right now we are busy with big refactoring tasks.

apoorvdarshan:
Rebased onto the current `master` and ported the new suite to the refactored `areg::String` API and naming (`length`, `as_string`, `find_first`, `areg::Radix`, etc.). The `areg-unit-tests` target builds successfully, and all 138 `StringTest*` tests pass.

aregtech:
@apoorvdarshan , thank you very much. I'll make a review and merge.

One immediate thing:
```
* \copyright   (c) 2017-2026 Aregtech UG. All rights reserved.
```
This is a little wrong, please remove the `UG`. If you think this file must be copyright protected (i don't think this is the case), you can protect with your name.

But one thing I should point -- set your name as the `author`. Having `Artak Avetyan` as the author is not fair, you are the author of these tests :)
I can be co-author if in the future add more tests

```
* \copyright   (c) 2017-2026 Aregtech. All rights reserved.
* \author      apoorvdarshan (your name or just your name)
```

apoorvdarshan:
Updated the String test header exactly as requested: removed UG from the Aregtech copyright and set the author to Apoorv Darshan. The areg-unit-tests target builds successfully and all 138 StringTest* tests pass.

aregtech (approved):

aregtech:
Thank you @aopv

PR #682, opened 2026-02-21
Collapse NE sub-namespaces into flat areg:: namespace

caliph1998:
## Summary

Collapses all `NE`-prefixed sub-namespaces (`NECommon`, `NEService`, `NERegistry`, etc.) into the flat `areg::` namespace by introducing `AREG_NAMESPACE_BEGIN/END/USE` macros, as part of the ongoing API modernization effort.

## Changes

- **Added namespace macros** to `framework/areg/base/GEMacros.h`:
  ```cpp
  #define AREG_NAMESPACE_BEGIN    namespace areg {
  #define AREG_NAMESPACE_END      }
  #define AREG_NAMESPACE_USE      using namespace areg;
  ```
- **Wrapped 16 NE-namespace Defs headers** in `AREG_NAMESPACE_BEGIN/END`, merging their contents into `areg::`:
  `NECommon`, `NEMath`, `NEMemory`, `NESocket`, `NEString`, `NEUtilities`, `NEDebug`, `NEApplication`, `NEService`, `NERegistry`, `NEConnection`, `NERemoteService`, `NELogging`, `NELogOptions`, `NEPersistence`, `NESyncTypesIX` (→ `areg::os::`)
- **Collapsed 3 library-specific namespaces** into their respective library namespaces:
  `NESystemService` → `aregext::`, `NELogCollectorSettings` → `logcollector::`, `NEMultitargetRouterSettings` → `mtrouter::`
- **Updated all call sites** across `framework/`, `examples/`, `tests/`, and `product/generate/` (~500 files, ~9300 reference replacements) from `NEXxx::` to `areg::`.

## Testing

```bash
cmake -B build -DAREG_BUILD_TESTS=ON -DAREG_BUILD_EXAMPLES=ON
cmake --build build -j20
ctest --test-dir build
```

All 359 unit tests pass. All 29 example targets build cleanly.

## Notes

- `namespace std` specialization blocks in class headers (e.g. `String.hpp`, `HashMap.hpp`) were intentionally left outside the `AREG_NAMESPACE_BEGIN/END` macros to avoid producing `areg::std::hash<T>`.
- `examples/23_pubdatarate/pubservice/` defines its own `namespace NEUtilities` for example-local types; those references were preserved and only the framework-originating symbols (`DataLiteral`, `convDataSize`, `convDuration`) were updated to `areg::`.
- Private OS-helper blocks in `SocketDefs.cpp` and `UtilityDefs.cpp` (previously `namespace NESocket`/`namespace NEUtilities`) were moved into `namespace areg` so the public functions in the same namespace can resolve them without qualification.

## Checklist

Please check the box ( `[x]` ) and sign-off the this PR:

- [x] I have read CONTRIBUTING.md

Signed-off-by: @caliph1998


aregtech:
Thank you, @caliph1998 
This looks good. 
I'll review

aregtech:
@caliph1998 There are many files that have conflict.
are you able to resolve conflicts or you need a help?

caliph1998:
@aregtech 

Thanks for reviewing. I will try to resolve the conflicts by tomorrow. It should be possible by just a rebase. If I can't do it by then, I will ask for your help. I appreciate it 😄

caliph1998:
@aregtech 

It seems one of the issues here is that I need to rebuild codegen.jar otherwise to make my PR work I'd have to add backwards-compatible aliases. Is that true? I might need some help with that it's not clear to me

aregtech:
@caliph1998 , right. I forgot that `codegen` also needs change :) 
No worries, if the problem is only `codegen.jar`, I can fix it. The main issue is to resolve conflict in these files:
```
examples/01_minimalrpc/src/main.cpp
examples/02_minimalipc/src/consumeripc.cpp
examples/03_helloservice/common/src/ClientComponent.cpp
examples/03_helloservice/common/src/ClientComponent.hpp
examples/07_logging/src/main.cpp
examples/08_timer/src/main.cpp
examples/10_sync/src/main.cpp
examples/11_service/src/ServicingComponent.hpp
examples/11_service/src/main.cpp
examples/12_svcmulti/src/ServicingComponent.hpp
examples/13_locsvc/locservice/src/ServiceClient.cpp
examples/13_locsvc/locservice/src/ServiceClient.hpp
examples/14_locmesh/locsvcmesh/src/ServiceClient.cpp
examples/14_locmesh/locsvcmesh/src/ServiceClient.hpp
examples/15_pubsvc/pubclient/src/ServiceClient.cpp
examples/15_pubsvc/pubclient/src/ServiceClient.hpp
examples/16_pubmesh/common/src/LocalHelloWorldClient.cpp
examples/16_pubmesh/common/src/LocalHelloWorldClient.hpp
examples/16_pubmesh/common/src/PublicHelloWorldClient.cpp
examples/16_pubmesh/common/src/PublicHelloWorldClient.hpp
examples/16_pubmesh/common/src/PublicHelloWorldService.cpp
examples/16_pubmesh/common/src/PublicHelloWorldService.hpp
examples/16_pubmesh/pubservice/src/PublicServiceComponent.cpp
examples/16_pubmesh/pubservice/src/PublicServiceComponent.hpp
examples/17_pubtraffic/pubclient/src/TrafficLightClient.cpp
```

Let me take this PR and fix `codegen.jar`

caliph1998:
@aregtech 

Thank you. I'm sorry I had a hard time with this Fix. It's a lot of file changes and details of these examples is unclear to me. A lot of them use their own separate namespace def files.

aregtech:
@caliph1998 , i tried to rebase your branch with branch 655, and have got ~330 conflict files :) 
What do you think if we drop this PR, create new branch in this repo and try to do the job together?
I'll also join to this task and in 1-3 days we'll finish the entire issue.

caliph1998:
@aregtech 

about those macros. I was going according to what's said in areg_modernization_plan.md:

### 2.4 Macro changes

| Macro | Old behavior | New behavior |
|-------|-------------|-------------|
| `AREG_NAMESPACE_BEGIN` | Conditional `namespace areg {` or nothing | Always `namespace areg {` |
| `AREG_NAMESPACE_END` | Conditional `}` or nothing | Always `}` |
| `AREG_NAMESPACE_USE` | Conditional `using namespace areg;` or nothing | Always `using namespace areg;` |
| `AREG_NAMESPACE` option | `0` or `1` | **Removed.** Always on. |

"_Keep the macros themselves — they provide consistent syntax across the codebase and allow future flexibility. Just remove the conditional behavior._"


The macros had no conditions they were always equal to using areg::. but they kept the option of adding a condition opened. Now that we are getting rid of macros altogether then we should also remove this from the plan markdown file.

aregtech:
@caliph1998 , ah, I see what you mean :) You are right. I forgot to update doc.
emmmm... let me explain why i removed macro.
First I was thinking that it is nice to have enable/disable `namspace areg`, but then, during implementation faced problems. For example, if I call a method of `areg` framework from `mtrouter`:
1. with no `namespace` it is a call `Application::loadModel()`;
2. with `namespace areg` it is a call `areg::Application::loadModel();`
3. with enabled/disabled flag there are 2 possibilities:
   a. either at the begin use `AREG_NAMESPACE_USE` and then go for plain `Application::loadModel()`;
   b. or use another macro, like `AREG_NS` (which means `areg`) and write a code `AREG_NS::Application::loadModel)`  

The solution (3a) is not good, because this limits calls from header files -- including `using namespace areg;` in header is never a good idea.
The solution (3b) is ugly. In addition, it would require additional macro like `AREGEXT_NS` / AREGLOG_NS, etc.. And all just to have this enable/disable macro, which in reality have no value.

Because of all these, i came to the conclusion that macro make the code less readable and ugly, which is definitely bad for learning path, and which is extremely important factor for any library.

In short, let's reduce macros as much as possible and instead use `namespace areg` :)

PR #687, opened 2026-03-02
Feature/616 modify ring stack for adding and inserting element

### LINKED ISSUE #616: Modify ring stack for adding and inserting element
**Description:**

When inserting a new element into the list, the function should return additional information indicating whether an existing element was removed during the operation — regardless of whether the insertion itself succeeded.

The return value should include:
1. The current number of elements in the list  
2. A flag indicating whether the element exists (or was successfully inserted)  
3. The removed element, if any  

Including a boolean flag is essential for validation. For example, if the list stores integers and only two fields are returned, it may be impossible to distinguish between a valid integer value and an indicator of removal. The explicit flag resolves this ambiguity.

The same logic should apply to the `remove()` method, which should also indicate the removed element.

This behavior can be introduced either by modifying the existing API or by adding a new method that provides extended return information.

collinshields:
## Summary
Modifies the push and pop methods to return a struct containing additional useful information.

Closes [#616](https://github.com/aregtech/areg-sdk/issues/616)

## Changes
- Added struct, `RSOperationResult`, containing the following information to `TERingStack.cpp`:
  1. `count` : The number of elements in the Ring Stack after a push/pop operation
  2. `isRemoved` : A flag that is true when an element is removed (shift overlap, or pop on non-empty Ring Stack)
  3. `removedElement` : The element that was removed, if any
- `push` and `pop` methods now return the above struct
- Modified existing tests to handle the new return struct
- Added additional checks to tests for new variables in the struct
- Updated line 121 of `NetTcpLogger.cpp` to handle the new return struct from pop

## Testing
Build:
```cmake -B .\build ` ```
```cmake --build .\build --config Release```
Test:
`.\product\build\msvc-cl\windows-64-x86_64-release-shared\bin\areg-unit-tests.exe --gtest_filter="TERingStackTest*"`

Project builds and all 14 TERingStack tests pass.
All 359 tests from the full testing suite still pass.

## Checklist

Signed-off-by: Collin Shields

### DIFF
--- a/framework/areg/base/TERingStack.hpp
+++ b/framework/areg/base/TERingStack.hpp
@@ -982,8 +982,11 @@ RSOperationResult<VALUE> TERingStack<VALUE>::push( const VALUE& newElement )
                 NEMemory::constructElems<VALUE>(block, 1);
                 *block = newElement;
                 ++ mElemCount;
+                return RSOperationResult<VALUE>(mElemCount, true, false, removedElement);
+            } else {
+                OUTPUT_WARN("Failed to resize Ring Stack in TERingStack::push() when overlapping is ResizeOnOverlap");
+                return RSOperationResult<VALUE>(mElemCount, false, false, removedElement);
             }
-            return RSOperationResult<VALUE>(mElemCount, true, false, removedElement);
 
         case NECommon::eRingOverlap::StopOnOverlap:
             OUTPUT_WARN("The new element is not set in Ring Stack, there is no more free space for new element");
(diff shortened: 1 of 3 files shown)


collinshields:
This is my first pull request, so let me know if there are any issues!

collinshields:
Ah, I just realized I messed up the signing. I've got it set up properly in VSCode now. Should I remake the branch and pull request? It should be easy enough to copy the changed files to a new branch.

aregtech:
@collinshields , just dropping a message to inform that I did not ignore your modification.
We'll come back to this topic after I merge the refactoring with the `master` branch, because it also changes the API.
Please give me a time.

collinshields:
It's no problem, I understand. Take your time!

aregtech:
The existing `RingStack` has been changed.
See `areg/base/RingStack.hpp`
I would appreciate very much if you optimize the existing ring stack
At the moment i'm closing this PR.

PR #695, opened 2026-03-21
Fixed performance under macOS. Used thread safe mFiredEntry variable

aregtech:
## Summary
Used mFiredEntry variable as atomic to escape data race, because 2 mutexes are used to read this data.

## Checklist

Signed-off-by: Artak Avetyan


vukihub (approved):
Looks fine to me.

PR #701, opened 2026-03-27
Feature/700 phase 3 optimize socket communication

aregtech:
## Summary
Multiple changes to optimize socket connection and communication. Integrated `epoll()`.

## Checklist

Signed-off-by: Artak Avetyan

### DIFF
--- a/framework/areg/base/private/posix/SocketMultiplexerPosix.cpp
+++ b/framework/areg/base/private/posix/SocketMultiplexerPosix.cpp
@@ -1,0 +1,452 @@
+/************************************************************************
+ * This file is part of the Areg SDK core engine.
+ * Areg SDK is dual-licensed under Free open source (Apache version 2.0
+ * License) and Commercial (with various pricing models) licenses, depending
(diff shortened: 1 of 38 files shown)


framework/areg/base/private/posix/SocketMultiplexerPosix.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+    const int32_t total = count + 1;
+
+    constexpr int32_t    STACK_MAX{ 32 };
+    struct pollfd        stackFds[STACK_MAX];
+    std::vector<struct pollfd> heapFds;
+    struct pollfd* const fds = (total <= STACK_MAX)
+                             ? stackFds
+                             : (heapFds.resize(static_cast<std::size_t>(total)), heapFds.data());
+
```
vukihub, line 416: Shouldn't we here check if sockets are more than max allowed? What if count is 100001?

vukihub (approved):
Looks good now.

PR #727, opened 2026-06-09
Feature/722 refactor event objects

### LINKED ISSUE #722: Refactor Event objects
**Describe:**
Currently, when messaging with `Event` the system makes more than one memory allocation -- one time for data, and second time for `Event` object. In addition, some events serialize `path` in the buffer and when deliver on target they are de-serialize the path to deliver event on the target. These all can be simplified if have a clean structure of Event header and event data. So that when event object is created it does not allocate 2-3 times, but only once similar to SharedBuffer or RemoteMessage.

aregtech:
## Summary
- Refactoring of Event object to reduce heap allocation;
- Optimized Event delivery to reduce latency;
- Cleaned up the code.

## Checklist

Signed-off-by: Artak Avetyan


vukihub (approved):
Too many files to review each of them separately.
Approving

PR #731, opened 2026-06-14
Feature/730 improve example 30

### LINKED ISSUE #730: Improve example 30
**Description:**
In order to make latency right measurements example 30 should be slightly optimized that the app is not spending time on variable initialization, rather on the performance measurement. The big data initialization takes time, so that it would be more senseful to use preinitialized data.

aregtech:
## Summary
- Improved example 30
- Found and fixed bug in timer and timer manager, in MpscMessageQueue

Signed-off-by: Artak Avetyan


vukihub (approved):
Looks fine

PR #739, opened 2026-06-17
Fixed bug of releasing per thread cache

### LINKED ISSUE #738: mtrouter memory not released after load
DESCRIPTION:
After high throughput test and stopping all clients/services, mtrouter keeps high memory usage (hundreds of MB) instead of returning near baseline.

STEPS TO REPRODUCE:

1. Start mtrouter, 23_pubservice, 23_pubclient
2. Run high load: c=30 -s
3. Stop service: in 23_pubservice use -q
4. Ensure all clients stopped
5. Check mtrouter memory usage

EXPECTED:
Memory returns close to initial baseline after all connections are closed.

ACTUAL:
mtrouter keeps significantly elevated memory usage even when idle.

IMPACT:
Possible memory leak or unreleased buffers under high load. Memory accumulates across test cycles.

NOTES:
Likely areas: message buffers, queues, pooling allocator, session cleanup, or fragmentation preventing OS release.

aregtech:
## Summary

- Fixed per thread cache bug: the cache is released when socket is not valid anymore. The batch array is released, then events are processed. This keeps mtrouter and logcollector to use less RAM when apps quit.
- Fixed but related with mutext lock -- some lock where missed.
- Added sanitizer tool to run

Signed-off-by: Artak Avetyan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 .idea
 
 build/*
+build-*/
 out/*
 product/*
 tools/*.zip
--- a/framework/areg/base/SharedBuffer.hpp
+++ b/framework/areg/base/SharedBuffer.hpp
@@ -520,7 +520,7 @@ protected:
      * \return  Returns the current writing position in the initialized buffer; returns
      *          INVALID_CURSOR_POSITION if the buffer is invalid.
      **/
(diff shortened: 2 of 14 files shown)


vukihub (approved):
Looks fine.

PR #747, opened 2026-06-27
Bugfixes: 746 fix examples

### LINKED ISSUE #746: Fix examples
**Describe the bug**
Not all examples are running.
For example, the `19_pubfsm` is broken. It should be fixed.

aregtech:
## Summary
- Fixed bug connecting logcollector;
- Fixed bug sending "hello server" handshake message;
- Fixed bug of sending messages before the connection is accepted -> first connection accepted, cookie is set, then send messages
- Fixed bug of sending logging scopes before messages -> before sending any message, send logging scopes so that the system knows the scopes and prios
- Fixed bug of reconnecting to the service when connection is lost,  made that on base class level, so that the logic is same for `logcollector` and `mtrouter`
- use `new` operator instead of calling `DEBUG_NEW` (which is `new(_NORMAL_BLOCK, __FILE__, __LINE__)`), because in MSVC it is aligning to 16-bit, which is causing segmentation fault -- in `EventQueue` we use 128-bit alignment.

Signed-off-by: Artak Avetyan


vukihub (approved):
Looks fine

PR #781, opened 2026-08-22
766 codegen adaption

### LINKED ISSUE #766: Codegen adaption
**Description:**

Adapted fixes for new code-generator.

The code generator has been updated:
- service interface designer;
- state-machine designer;
- data validator.

Make accumulative changes of areg-sdk to adapt new codegen.

aregtech:
## Summary

Consumer hangs on macOS
Merging PR from previous build shown that the tests were crashed, regardless the complete status was green
Deeper analyze demonstrate that there is a bug to fix.

## Changes

Removed the `EV_DELETE`-on-`EV_EOF` block from **both** the cached-serve path and the
first-event path, and documented the socket-lifetime contract in the design comment.
A socket now stays registered until `unregister_socket()` or `reset()` removes it --
identical to the generic `poll` backend. Also dropped the two locals that became unused.

## Testing

Run all examples, everything is green.

## Notes

The macOS kqueue backend of `SocketMultiplexer` removed a socket from the kqueue the
moment it observed `EV_EOF`, **without checking whether readable bytes were still
buffered**. kqueue sets `EV_EOF` together with `data > 0`, so under load the provider's
final EOF arrives glued to a backlog of unread messages. The router consumed that one
delivery as an ordinary successful read, never ran its cleanup path, and the socket was
then invisible to the kqueue forever. The router therefore never learned that the
service provider had gone away, never sent the unregister notification, and the consumer
parked in `Application::wait_quit()` until the harness killed it.

Linux (`epoll`) deletes only when there is no readable data; the generic `poll` backend
never deletes. **macOS was the only unsound backend**, which is exactly why the defect
was macOS-only.

## Checklist

Signed-off-by: Artak Avetyan


vukihub (approved):
looks fine. a little bit too much comments.

PR #785, opened 2026-08-26
Made event queue and threads more embedded friendly, saved ~240KB space.

### LINKED ISSUE #784: Make the framework embedded friendly
## Describe

Each dispatcher thread allocates 1024 x 56 bytes of message queue when initialized.
On an empty start, areg based apps start a few threads, some of them might even never run. For example, if apps do not use Watchdog or Timer, there is no need to start Watchdog- and Timer-manager threads, but since buffers are allocated in the constructor, if fill cost `2 x 56 x 1024` bytes for nothing.

In addition, the threads like ServiceManager do not require large queue size. Large queues are senseful in the threads like socket send / receive, because in case of IPC we should make sure that there is enough space allocated for messages send/receive from multiple components. But this particular part is configurable in `areg.init` file.

## Agenda

Make areg framework more embedded friendly. Check if the `EventQueue` is optimized and does not burn CPU, make sure that message queues are created by need.

aregtech:
## Summary

Reduced the event queue size for some threads.
Fixed looping / CPU burn in EventQueue

## Checklist

Signed-off-by: Artak Avetyan

### DIFF
--- a/conf/cmake/functions.cmake
+++ b/conf/cmake/functions.cmake
@@ -603,11 +603,16 @@ macro(macro_system_bitness var_bitness)
             set(${var_bitness} ${_bitness_value})
         else()
             set(${var_bitness} 64)
+            # Called before 'project()', none of the three sources exists yet and the
+            # value is a placeholder that 'common.cmake' replaces once the compiler is
+            # known. Report only a real failure: a known host that stays unresolved.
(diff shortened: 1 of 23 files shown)


vukihub (approved):
tested on linux. no regression

PR #792, opened 2026-09-02
tests(base): Add unit tests for std::hash specialization and required_size templates

Kuba70:
## Description
This PR increases unit test coverage in `TEString.hpp` by adding tests for custom template specializations and helper structs.

## Summary of Changes
- Added `StdHashSpecialization` test to verify `std::hash<areg::String>` behavior (including empty strings, spaces, and case-sensitivity).
- Added `RequiredSize_StdStringView` test to verify serialization buffer size calculation for `std::string_view`.
- Added `RequiredSize_StdString` test to verify buffer size calculation for `std::string`.
- Added `RequiredSize_aregString` test to verify buffer size calculation for `areg::String`.

## Testing
- All tests pass locally using Google Test framework (`StringTest` suite).

## Checklist

Signed-off-by: Kuba70 / Kuba Słowikowski


aregtech:
@Kuba70 , thank you very much.

Can you please fix the PR? You need copy and paste this section in the PR that the validation checker passes:

```
## Checklist

Signed-off-by: Your Name
```

In this `Your Name` you should have either your full name, or GitHub nick-name. Ideally, to have both

aregtech (approved):
Looks OK

aregtech:
@Kuba70, thank you very much for your contribution!
I’d really appreciate it if you could also add tests for more complex features and methods.

I’d also like to add benchmark tests, like [this one](https://github.com/aregtech/areg-sdk/pull/783/changes), and I’d be happy if someone would be interested in managing and maintaining this part of the project.

Strings, synchronization objects, containers, etc. are used a lot, so I’d like to benchmark their performance. It would also be interesting to measure things like how quickly a thread starts and shuts down, how fast configuration files are read and parsed, and so on.

Most importantly, I’d like to identify code paths, nodes, or calls that cause high latency.

Benchmarking and optimization are serious tasks, but since there is no top urgency for any of this, you can do that relaxed, whenever you have time :)

Kuba70:
Thanks, Artak! I'm really excited about this opportunity and would be glad to take care of benchmarking and performance in areg-sdk. I'll work on this consistently at a steady pace. Thanks for the trust!

Kuba70:
Hi @aregtech ,

I’ve been going through the existing benchmark suite for DateTime and String and noticed that most core operations are already implemented.

Before adding more tests, I’d like to align on the project's strategy for benchmarking. How would you prefer I proceed?

Complete coverage: Add micro-benchmarks for all small remaining public methods.
Targeted bottleneck analysis: Focus specifically on high-overhead paths, such as heap allocations, complex text parsing, and time formatting/conversions.

Expand scope: Move on to setting up benchmarks for other uncovered modules in the repository.

Looking forward to your feedback so I can align my PR with the project goals!

aregtech:
Hi @Kuba70 , 
You ask very good question.
I was thinking to integrate [CodSpeed](https://github.com/aregtech/areg-sdk/pull/783). This is a PR created by their agent.
Honestly saying, I don't have yet a clear strategy. The agenda should be measurable performance. Similar to unit tests to catch nodes that cause latency.
Feel free to make suggestions. I've had quick view on your other PR -- do you think I should use `google_benchmark`? Is this your recommendation?

aregtech:
I think we can open a topic for discussion.

Kuba70:
Hi @aregtech , 

Thanks for the reply! I’m happy to help develop the performance testing strategy; I did a lot of reading about benchmarks over the weekend, and regarding your question about comparing Google Benchmark and CodSpeed:

Google Benchmark is the industry standard for C++ micro-benchmarks. CodSpeed ​​is actually built upon Google Benchmark—it uses its API to measure CPU instruction counts and detect performance regressions within the CI process.

In my latest PR, I updated the `CMakeLists.txt` file to include a `CODSPEED_MODE` switch (set to `OFF` by default). This allows us to combine the benefits of both solutions:

Developers can run benchmarks locally using standard Google Benchmark without needing to install specialized CodSpeed ​​dependencies.

In the CI/CD pipeline, `CODSPEED_MODE=ON` can be enabled to automatically track performance regressions on GitHub.

I think it would be a good idea to start a separate thread for performance testing so we don't mix topics.

aregtech:
Hi @Kuba70 ,
sorry for late reply. I was busy with some very important experiments and did have much time :)

OK. Thank you! I think it is reasonable. Then I can accept [this PR](https://github.com/aregtech/areg-sdk/pull/783) and your benchmark change on top of that.

P.S. I would like to move our talk to the `discussion`. Interesting information, which should not be lost.

Kuba70:
No problem at all, Artak! Thanks for the update.

That sounds like a great plan. Feel free to merge the PR and start a new thread in Discussions whenever you're ready. I'll be happy to join the conversation there and share more detailed ideas on how we can structure the benchmark suite and CI integration going forward.

Looking forward to it

